PR #1, opened 2026-07-02
Code-review fixes: rollout forensics, eval-log survival, lifecycle & validation hardening

aris-zhu:
## Summary

Fixes every finding from the full codebase review, in 7 focused commits. All gates green: ruff, ruff format, mypy strict (now covering `tests/` too), pytest at 100% line+branch coverage (143 tests).

### Correctness
- **Never lose the log**: reducer failures (`pass_at_k` over fewer epochs than k after a halt, `mean` over categorical scores) degrade to an error `EvalLog` instead of crashing `eval()` on the exact failure path it exists to survive; unknown reducers fail fast as `ConfigError`.
- **Errored trials are never scored**: a `PolicyError` trial can no longer poison metrics (e.g. `min_distance_to_goal`'s `inf` on an empty record). Failures stay visible via `SceneResult.status` and an empty `epochs` entry.
- **Forensics survive errors**: every error raised inside a trial carries the partial `TrialRecord` (steps walked + transcript events) on `exc.record`; `eval()` counts it and delivers it to sinks — for continue-class and halt-class failures alike.
- **Correct attribution**: `reset()` failures are wrapped into the taxonomy; a crashing approver becomes `SafetyAbort`; a policy emitting wrong-dimension actions is a `PolicyError`, not an eval-halting `EmbodimentFault` blamed on the robot.
- **Lifecycle**: `eval()` closes embodiments it resolved from registry names — even on a halt or compat failure — so a real robot never leaks its connection; caller-constructed objects stay caller-owned.
- **Seeding**: `seed=None` draws recorded OS entropy instead of silently behaving as `seed=0`; `derive_seed` no longer aliases `None`/`0`.
- **Transcript fidelity**: `DefaultController` records the actual buffered action count; `ClampApprover` modifications now produce the documented approval events.

### Design / robustness
- `fail_on_error` is checked after every trial, so `True` stops at the first error.
- `RerunSink` namespaces entities per trial (trials no longer overwrite each other), warns per instance, drops the dead `_t` counter, and supports both old and >=0.23 rerun-sdk API surfaces.
- Registry entry-point discovery warns on broken plugins instead of swallowing them.
- `Task`/`Epochs` validate at construction (`ConfigError`); `Task.scorer` accepts registry names without the str-as-Sequence footgun; `Box`/`ObservationSpace` validate bounds and state-key consistency.
- `FrameStore` slugs are collision-safe; `_git_commit()` appends `-dirty` for unclean trees; CLI prints the written log path and gains `--epochs`/`--fail-on-error`/`--store-frames`; dead code removed (`JsonLogSink._steps`, coverage-driven placeholder statements in `rollout`).

### Tooling / docs
- mypy strict now covers `tests/`; CLAUDE.md documents the full CI matrix story and the new error-handling invariants.

Note: trial seeds differ from previous runs for `seed=None`/`seed=0` (the aliasing fix) — intentional, pre-1.0.

## Testing
- `ruff check .` / `ruff format --check .` clean
- `mypy` strict: 43 files, no issues (src + tests)
- `pytest --cov`: 143 passed, 100.00% coverage (gate: `--cov-fail-under=100`)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,11 +29,13 @@ rollout, scores it, and writes an immutable `EvalLog`. Mirrors Inspect AI's
 
 - Dev loop: `uv pip install -e ".[dev]"`, `uv run pre-commit install`, then
   `uv run pytest --cov`.
-- Gates that must pass: `ruff check .`, `ruff format --check .`, `mypy` (strict),
-  and `pytest --cov` at **100% coverage** (`--cov-fail-under=100`). Pre-commit
-  runs ruff+mypy on commit and the coverage gate on push
-  (`.pre-commit-config.yaml`). CI runs all gates on Linux+macOS / py3.11-3.12 as
-  **blocking, required PR checks**; coverage below 100% fails the build.
+- Gates that must pass: `ruff check .`, `ruff format --check .`, `mypy` (strict,
+  covers `src` **and** `tests`), and `pytest --cov` at **100% coverage**
+  (`--cov-fail-under=100`). Pre-commit runs ruff+mypy on commit and the coverage
+  gate on push (`.pre-commit-config.yaml`). CI runs all gates on Linux+macOS /
+  py3.11-3.12 as **blocking, required PR checks** (coverage below 100% fails the
+  build), plus a **non-blocking** `test-extra` tier for py3.10/py3.13, Windows,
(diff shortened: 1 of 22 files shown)


jeqcho:
Superseded by #9: forward-ported through the roboinspect→inspect-robots rename onto current main.

PR #15, opened 2026-07-07
isaacsim: pass parse_env_cfg result to gym.make (env creation was broken)

jeqcho:
## Summary

Three fixes that make the isaacsim plugin actually work on a live Isaac Lab install. Found and verified end-to-end on an RTX 5090 (Isaac Sim 5.0.0/5.1.0, Isaac Lab 2.2.0/2.3.2, `Isaac-Lift-Cube-Franka-v0`) while exercising the zero-config `--sim` path (#14).

1. **Env creation was unconditionally broken.** `_ensure_env` called `gym.make(self.task_id, num_envs=1, render_mode="rgb_array")` — but every Isaac Lab env takes a **mandatory `cfg` object**, so real env creation always failed with:

   ```
   EmbodimentFault: ManagerBasedRLEnv.__init__() missing 1 required positional argument: 'cfg'
   ```

   Fix: resolve the config with Isaac Lab's own `parse_env_cfg(task_id, device, num_envs=1)` and pass it as `cfg=`. (The plugin's CI tests run without Isaac by design — they inject a fake env — so only a live GPU run could catch this.)

2. **Named observation terms.** Manager-based envs concatenate a group's terms into one flat tensor by default (`concatenate_terms=True`), which leaves `_to_observation` with nothing to map onto the adapter's *named* state fields — policies would see an empty `Observation.state`. `_request_named_obs_terms` flips the group's `concatenate_terms` to `False`, and warns if it can't (missing group / renamed flag), so a silently-flat group is at least loud.

3. **Headless runs disable `debug_vis`.** Debug-visualization markers exist for a human watching a viewport — a headless eval has none. Spawning their prototype prims also drags in the shader/material machinery (`Sdr.Registry()` / MDL discovery), which can hang env creation outright on hosts whose render stack is unhealthy (observed live: a missing `libGLU.so.1` killed `omni.iray.libs` and the marker's preview-surface material spun forever). `_disable_debug_vis` walks the resolved env cfg (cycle-safe) and switches every `debug_vis=True` off.

Silver lining from the original failed run: the framework's error contract held — the `EmbodimentFault` halted the eval and the error `EvalLog` still reached disk with the fault recorded on the scene.

## Test plan

- Plugin suite (no-Isaac, fake-env): 15 passed, including new tests for the debug-vis walk (nesting, cycles, non-bool sentinels) and the named-obs request (success path quiet, no-op path warns); ruff + strict mypy (CI's exact invocation) green.
- Live verification on the 5090 box: **passed** on CPU physics (Isaac Sim 5.0.0) and GPU physics (5.1.0) — see comments below.
- `CHANGELOG.md` entry under "Unreleased".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,15 @@ All notable changes to this project are documented here. The format is based on
 
 ### Fixed
 
+- **isaacsim plugin: real env creation was broken.** `_ensure_env` called
+  `gym.make(task_id)` without the mandatory Isaac Lab `cfg` object, so every
+  live run failed with `missing 1 required positional argument: 'cfg'`; the
+  config is now resolved via Isaac Lab's own `parse_env_cfg`. Alongside it:
+  observation groups are requested as *named* dicts (`concatenate_terms=False`
+  — a flat tensor left `Observation.state` empty; a warning fires when the
+  request can't be honored), and headless runs disable every `debug_vis` flag
+  (markers exist for a viewport nobody has, and their material machinery can
+  hang env creation on hosts with a broken render stack).
 - **Eval logs are strict RFC 8259 JSON.** Non-finite floats (e.g. an inf
   `min_distance_to_goal` when no distance was ever recorded) are mapped to
   `null` at the JSON boundary, so `jq` and other conforming parsers accept the
--- a/plugins/inspect-robots-isaacsim/src/inspect_robots_isaacsim/embodiment.py
+++ b/plugins/inspect-robots-isaacsim/src/inspect_robots_isaacsim/embodiment.py
@@ -208,8 +208,13 @@ class IsaacSimEmbodiment:
         # Importing the tasks registers the gym ids; must happen AFTER the app boots.
         import gymnasium as gym
         import isaaclab_tasks  # noqa: F401  (registers Isaac-* gym ids)
+        from isaaclab_tasks.utils import parse_env_cfg
 
-        self._env = gym.make(self.task_id, num_envs=1, render_mode="rgb_array")
+        # Isaac Lab envs take a mandatory cfg object (gym.make(task_id) alone
+        # raises "missing 1 required positional argument: 'cfg'"); parse_env_cfg
+        # is Isaac Lab's own task-id -> config resolution.
+        env_cfg = parse_env_cfg(self.task_id, device=self.device, num_envs=1)
+        self._env = gym.make(self.task_id, cfg=env_cfg, render_mode="rgb_array")
         return self._env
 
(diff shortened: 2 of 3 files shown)


jeqcho:
**Live verification on real hardware: PASSED** ✅

Setup: `omen` (RTX 5090, driver 580, Isaac Sim 5.0.0 + Isaac Lab 2.2.0), `Isaac-Lift-Cube-Franka-v0`, zero-config invocation with a config-file default policy (`random8`, an 8-DoF entry-point test plugin) and `sim_embodiment = isaacsim`:

```
$ inspect-robots "lift the red cube" --sim --no-prompt --max-steps 60 -E device=cpu
policy: random8 (from ~/.config/inspect-robots/config.ini)
embodiment: isaacsim (--sim, from ~/.config/inspect-robots/config.ini)
status: success
scenes: 1  trials: 1
  success_at_end: 0
log: /tmp/ir-logs/adhoc_9917e246.json
```

Trial-level: scene `success`, **60/60 steps executed**, `metrics: {'success_at_end': 0.0}` (honest score — a jitter policy doesn't lift cubes), instruction + judgement slots persisted in the log. Negative control also passed: same command without `--sim` refuses with `no embodiment named 'missing-real-arm'`.

Each of the three commits fixed a failure observed live, in sequence:
1. `parse_env_cfg` → killed `ManagerBasedRLEnv.__init__() missing 'cfg'`
2. headless `debug_vis` off → killed an indefinite hang in `Sdr.Registry()` shader discovery during goal-marker creation (root-caused via SIGUSR2 faulthandler stack samples; Kit's breakpad handler eats SIGABRT)
3. `concatenate_terms=False` → killed `PolicyError: 'joint_pos'` (flat obs tensor vs the adapter's declared named state fields)

Notes for the record: physics ran on CPU because Isaac Sim 5.0's vendored `omni.warp` fails CUDA init on driver 580 (CUDA 13 dropped a legacy entry point) — box-level issue, tracked separately; pip `warp-lang` 1.16 works fine on the same driver.

jeqcho:
**GPU-physics addendum**: after upgrading the box to Isaac Sim **5.1.0** + Isaac Lab **2.3.2** (fresh venv), the vendored-Warp CUDA-init failure on driver 580 is gone (0 Warp errors at boot), and the same zero-config eval passes on **GPU physics** — no `device=cpu` override:

```
status: success · scene: success · 60/60 steps · 7.8s · metrics: {'success_at_end': 0.0}
```

So this PR's three fixes are verified against both Isaac Sim 5.0 (CPU physics) and 5.1 (GPU physics), and the adapter works across Isaac Lab 2.2 → 2.3 unchanged.

PR #23, opened 2026-07-07
CI: test the rerun extra for real; add format check + coverage to plu…

### LINKED ISSUE #6: CI: no job exercises the rerun extra; plugin job lacks format check and coverage
## Problems

1. **rerun extra never tested in CI:** no job installs `.[rerun]`, so `tests/test_rerun_sink.py`'s real-SDK test always skips. The sink calls SDK APIs against an uncapped `rerun-sdk>=0.20` floor, so upstream deprecations/removals ship silently (this already happened once — fixed on the `review-fixes` branch, being forward-ported). A small ubuntu leg installing the rerun extra and running the sink tests would catch drift.
2. **Plugin CI job** (`.github/workflows/ci.yml`, plugin job) runs `ruff check` but not `ruff format --check`, and plugin tests run without coverage measurement — the core enforces 100%, the plugin enforces nothing.

Tushar2604:
## Summary

RerunSink was only ever exercised against a fake `rerun` module in `test_coverage_completion.py`, so a real `rerun-sdk` API change would go unnoticed. Also, the isaacsim plugin CI job had lint but not format checking, and ran its tests with no coverage measurement at all, unlike the core's 100% gate.

## Changes

- Add a blocking `test-rerun` CI job that installs the real, uncapped `rerun-sdk>=0.20` floor and runs `test_rerun_sink.py` against it (previously always skipped in CI).
- Add `test-rerun` to `ci-ok`'s and `alert-red-main`'s `needs` lists so it actually gates merges.
- Add `ruff format --check` to the `plugin-isaacsim` job (it only ran `ruff check`).
- Add coverage reporting to the plugin's pytest run (`--cov=inspect_robots_isaacsim --cov-report=term-missing --cov-fail-under=0`) — report-only, not gated, since the plugin's `reset()`/`step()` paths need real Isaac Lab, which CI doesn't have.
- `CHANGELOG.md` entry under "Unreleased".

## Checklist

- [x] Tests added/updated (and the `CubePick` mock exercises the change where applicable) — N/A code change; verified the new CI steps locally instead (installed real `rerun-sdk`, confirmed `test_rerun_sink_writes_recording` passes; ran plugin coverage locally)
- [x] `mypy` passes (strict)
- [x] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test — N/A, no public API change
- [x] Core stays NumPy-only (new deps are optional extras, lazily imported) — unaffected, CI-only change

## Related

Closes #6


jeqcho (commented):
Thanks for this contribution, @Tushar2604 — and welcome! Auditing CI for gaps like this is genuinely valuable work: you're right that `RerunSink` was only ever exercised against a fake `rerun` module, and that the plugin job was missing a format check and any coverage visibility. The PR also gets several repo-specific conventions right that are easy to miss: the new `test-rerun` job was added to both `ci-ok.needs` and `alert-red-main.needs` (per CLAUDE.md, jobs not in `ci-ok.needs` don't gate merges), it keeps `uv sync --locked`, keeps `fetch-depth: 0` for hatch-vcs, and `--cov-fail-under=0` correctly keeps the plugin out of the core 100% gate ("plugins never count toward the core gate") — I verified the root `pyproject.toml`'s `[tool.coverage.report] fail_under = 100` really would apply otherwise, so that override is load-bearing, and your comment explaining it is appreciated.

I verified the new job commands empirically on your branch (macOS, uv):

- `uv sync --locked --python 3.12 --extra dev --extra rerun` — succeeds; installs the locked `rerun-sdk==0.34.0` (the `rerun` extra is already in `uv.lock`, so `--locked` is happy).
- `uv run pytest tests/test_rerun_sink.py -v` — **2 passed, 2 skipped**: `test_rerun_sink_writes_recording` now actually runs against the real SDK (previously always skipped in CI), and the two "absent path" tests correctly skip. Exactly the intent of #6.
- `uv run --no-sync ruff format --check plugins/inspect-robots-isaacsim` — passes ("4 files already formatted").
- `uv run --no-sync python -m pytest plugins/inspect-robots-isaacsim/tests -q --cov=inspect_robots_isaacsim --cov-report=term-missing --cov-fail-under=0` — 13 passed, reports 75% coverage, does not fail. Works as described.

No workflow security concerns: the new `run:` blocks contain no untrusted `${{ }}` interpolation.

Findings:

1. **[blocker] Merge conflict with main in `.github/workflows/ci.yml`.** GitHub reports the PR as CONFLICTING (I reproduced it locally: the conflict is only in `ci.yml`; `CHANGELOG.md` auto-merges). After this PR was opened, main gained a `plugin-xpolicylab` job (`ci.yml:162-186` on main) and both `needs` lists were extended to include it (`ci.yml:193` and `:206` on main) — the exact lines your PR also edits. Please rebase onto main and resolve so the final `needs` lists contain **both** additions: `[quality, test, test-rerun, test-extra, core-only-import, plugin-isaacsim, plugin-xpolicylab]` (same for `alert-red-main`). Dropping `plugin-xpolicylab` in the resolution would silently un-gate it, so worth double-checking after the rebase. This is also almost certainly why no CI checks have run on this PR yet — they should kick off once it's mergeable.

2. **[suggestion] Apply the same treatment to the new `plugin-xpolicylab` job when you rebase.** It was copied from the old shape of `plugin-isaacsim`, so it has the same two gaps this PR fixes: `ruff check` without `ruff format --check`, and pytest with no coverage reporting. Mirroring your `Ruff (plugin, format check)` step and the report-only `--cov=inspect_robots_xpolicylab --cov-report=term-missing --cov-fail-under=0` pytest invocation there would complete the audit in one PR (its tests use an in-process stub server, so coverage may well be high enough to be interesting).

3. **[nit] Comment/changelog wording: "installs the real, uncapped rerun-sdk>=0.20 floor".** Since the job (correctly, per the "CI installs from uv.lock" convention) uses `uv sync --locked`, it installs the *pinned lock version* (0.34.0 today) — not the `>=0.20` floor and not the latest release. The weekly canary is what exercises the newest versions the ranges allow. Suggest rewording the `ci.yml` comment (around the new `test-rerun` block) and the PR/CHANGELOG text to "installs the locked rerun-sdk" so a future reader doesn't expect floor/latest testing from this job. Behavior is fine as-is; only the description overstates it.

Once the rebase lands and CI goes green, this looks good to go from my side — the core idea (a blocking `test-rerun` job plus report-only plugin coverage) fits the repo's CI philosophy well.

🤖 This review was performed by [Claude Code](https://claude.com/claude-code) on behalf of the maintainer.

jeqcho:
Thanks @Tushar2604 — two solid PRs in one go! 🎉 This one caught a real blind spot: `RerunSink` was only ever tested against a fake `rerun` module, so a real SDK break would have sailed through CI. The `--cov-fail-under=0` detail (overriding the root `fail_under=100` that coverage.py would otherwise inherit) was a particularly careful touch. We mirrored your new format/coverage steps onto the `plugin-xpolicylab` job that merged in the meantime, and everything is green. Merging!

Hope to see you around here again — issues are open and PRs like these are exactly what we want more of. Welcome aboard! 🤖

Tushar2604:
@jeqcho Thanks a lot! Really appreciate the review. Glad the fix was helpful—looking forward to contributing more! 🚀

PR #30, opened 2026-07-10
CLI closes resolved embodiments; store_frames config default + per-run frame dirs

aris-zhu:
Two hardening changes for real-hardware runs, stacked (second builds on the first):

**1. CLI closes the embodiment it resolved** — `_cmd_run` resolves the embodiment into an object, so `eval()`'s "close what we open" ownership rule never applied and nothing closed it. Harmless for sims, but a real-hardware embodiment releases motor torque (and parks the arms) in `close()`; skipping it leaves the robot energized after the run. The close lives in a `finally` so a halting eval can't skip it. Found driving real YAM arms via the zero-config CLI.

**2. `store_frames` as a config default + per-run frame directories** — `[defaults] store_frames = true` in config.ini enables frame capture on every run without the flag. Frame side-cars now land in a per-run subdirectory of `<log_dir>/frames`: trial ids repeat across runs (scene-epoch), so the old shared directory silently overwrote the previous run's frames — unacceptable once capture is a default. `stats.frames_dir` records the exact directory.

Both TDD'd; ruff/mypy-strict/pytest green, coverage untouched.

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -39,6 +39,7 @@ embodiment = yam-bimanual            ; the default: real hardware
 sim_embodiment = yam-bimanual-isaac  ; what --sim swaps in
 scorer = operator      ; optional, ad-hoc runs only
 max_steps = 300        ; optional, ad-hoc runs only
+store_frames = true    ; optional, capture frames on every run
 
 [policy.args]          ; default -P key=value pairs
 checkpoint = ~/ckpts/molmoact2-yam.pt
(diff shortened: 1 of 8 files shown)


jeqcho:
Pre-merge fresh-eyes review loop (3 independent subagent rounds):

- **Round 1**: `--store-frames` was a plain `store_true`, so a `store_frames = true` config default couldn't be disabled per-invocation → made it a tri-state `BooleanOptionalAction` (`--no-store-frames` now overrides config); also fixed docs that still claimed frames land directly in `<log-dir>/frames` and added the new config key to the config reference.
- **Round 2**: the close-embodiment `try/finally` started too late — `--epochs 0` (ConfigError) between resolution and `eval()` leaked an energized embodiment → widened the span to start right after resolution, with a regression test.
- **Round 3**: clean; verified no double-close, no unbound `logs`, no behavior change from the restructure.

All gates green each round (ruff, format, mypy strict, pytest at 100% coverage).

PR #31, opened 2026-07-10
isaacsim: headless debug_vis + named observation terms

aris-zhu:
Two isaacsim plugin fixes that were stranded on a local branch when #15 (env creation) was merged independently:

- **Disable `debug_vis` in headless env configs** — headless runs crashed/warned trying to build debug visualization.
- **Request named observation terms (`concatenate_terms=False`)** — the embodiment needs per-term observations rather than one concatenated tensor.


aris-zhu:
Both commits turn out to already be on main (the earlier 'missing upstream' check was against a truncated path). Nothing to merge.

PR #41, opened 2026-07-10
fix: success evaluation fallback, handle empty/dark float frames, and…

### LINKED ISSUE #5: isaacsim plugin: 'terminated implies success' fallback can score failures as successes
## Problems (plugins/inspect-robots-isaacsim/src/.../embodiment.py)

1. **`_read_success` fallback (embodiment.py:289-293):** when the Isaac Lab task exposes no `success` info key, bare `terminated` is treated as success. Isaac Lab tasks also terminate on *failure* (dropped object, out of bounds, illegal state), so `success_at_end` can score failures as successes on any task without the oracle key. At minimum this deserves a loud warning when the fallback engages; arguably it should score as not-successful and require an explicit opt-in.
2. **`import gymnasium as gym` in `_ensure_env` (embodiment.py:209)** is not wrapped by `_missing_isaac`, and `gymnasium` is not declared in the plugin's dependencies — with torch installed but Isaac/gymnasium absent, users get a raw ImportError instead of the curated RuntimeError.
3. **`_to_image` float heuristic (embodiment.py:324):** `arr * 255.0 if arr.max() <= 1.0 else arr` misfires on a legitimately dark 0-255-range float frame (all pixels ≤ 1.0 get multiplied again), and `arr.max()` on an empty array raises.

Amresh-01:
## Summary
Fixes three issues in the Isaac Sim embodiment adapter:
1. `_read_success` fallback now warns and defaults to unsuccessful,
   with explicit `terminated_implies_success` opt-in
2. Imports in `_ensure_env` wrapped in try/except with proper error messaging
3. `_to_image` handles empty arrays and dark 0-255 float frames correctly

## Changes
- Added `terminated_implies_success` parameter to `IsaacSimEmbodiment`
- `_read_success`: warns on missing oracle key, returns False by default
- `_to_image`: guards against empty arrays and dark float frames
- `_ensure_env`: imports wrapped in try/except
- Added `gymnasium` to plugin dependencies
- Added comprehensive tests for all edge cases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,14 +27,19 @@ jobs:
           cache-dependency-glob: |
             **/pyproject.toml
             **/uv.lock
-      - name: Install from uv.lock (dev extras)
-        run: uv sync --locked --python 3.11 --extra dev
+      - name: Install from uv.lock (dev + docs extras)
+        run: uv sync --locked --python 3.11 --extra dev --extra docs
       - name: Ruff (lint)
         run: uv run ruff check .
       - name: Ruff (format check)
         run: uv run ruff format --check .
       - name: Mypy (strict)
         run: uv run mypy
+      # Gated here because docs.yml's build job is in a different workflow and
+      # can't join ci-ok's needs list: a dead link that fails the strict docs
+      # build must not be mergeable.
+      - name: Docs build (strict)
+        run: uv run mkdocs build --strict
 
   # Blocking test gate across the supported OS x Python matrix.
   test:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 
+- Plugin-declared embodiment device slots for V4L2 cameras, SocketCAN
+  interfaces, and serial devices. `inspect-robots setup` probes and interviews
+  declared slots, enforces grouped all-or-none assignments, and suggests udev
+  serial pinning for order-dependent USB-CAN names (#61).
+- Runtime-requirement declarations for registered component factories, with
+  missing-import preflight checklists in `inspect-robots setup` and
+  `inspect-robots doctor` (#59).
 - isaacsim plugin: `_ensure_env`'s cfg-wiring contract (`parse_env_cfg`'s
   args, `gym.make(cfg=...)`, the `headless` → `_disable_debug_vis` gate, and
   the named-obs-terms request) is now exercised in CI via stubbed
@@ -28,6 +35,16 @@ All notable changes to this project are documented here. The format is based on
 
 ### Fixed
 
+- **Literal percent signs in config values now round-trip unchanged** (#54).
+  Config reads no longer treat `%` as interpolation syntax, so values such as
+  `policy = 50%off` work with `config set`, `config show`, and normal runs.
+- **Component argument mistakes now fail cleanly and stale args are flagged**
+  (#47). Changing a configured component warns when its non-empty args section
+  still belongs to the old name, and invalid constructor kwargs exit with
+  guidance to check the config section or CLI args flag instead of a traceback.
+- **`inspect-robots run` now surfaces evaluation failures in its summary**:
+  top-level errors, per-scene failure context, and a ready-to-run postmortem
+  `inspect` command are printed after unsuccessful runs (#57).
 - **Config `[*.args]` sections no longer follow a differently-selected
   component** (#44). `[policy.args]` / `[embodiment.args]` /
   `[sim_embodiment.args]` now apply only when the selected component matches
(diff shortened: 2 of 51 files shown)


jeqcho:
Thanks for taking this on. The `_read_success` opt-in with the one-time warning and the wrapped `_ensure_env` imports are what #5 asked for, and the coverage on those paths is solid.

The `_to_image` change needs another pass though. A real dark 0-255 float frame almost always has fractional pixel values (0.3, 0.7, ...), so `has_fractional` is true and it still gets scaled by 255. Meanwhile a normalized frame containing only 0.0 and 1.0 values (a black frame with pure white pixels, say a float mask stacked to RGB) is no longer scaled, so white comes out as uint8 `1` instead of `255`. That's a regression from the old behavior. I verified both cases numerically. The underlying problem is that the range is ambiguous from values alone, so I'd keep only the empty-array guard from this part and either document "floats are treated as normalized 0-1" or add an explicit range knob on the constructor, rather than a smarter guess.

Two smaller things: `terminated_implies_success` should be documented in the constructor docstring and the README config table (including the scoring-default change, since tasks without the oracle key now score terminated episodes as failures), and please drop or version-bound the new `gymnasium` dependency. Isaac Lab pins its own, and the wrapped import already produces the curated error message.

Amresh-01:
Hi @jeqcho thanks for the review! I’ve documented `terminated_implies_success`, dropped `gymnasium`, and resolved the changelog conflict. For `_to_image` I’ll remove the heuristic and keep only the empty-array guard. Quick preference: assume all floats are 0–1 (always scale by 255), or add a `float_image_range` knob? Happy to go with your call.

jeqcho:
Let's go with assuming all floats are 0–1 (always `clip(0, 1) * 255`), no knob.

Reasoning: 0–1 is the convention for float images — 0–255 floats aren't a competing format, they're almost always an upstream bug (a `.float()` cast on a uint8 tensor with the `/255` forgotten). Always-scale handles all well-formed input, and on that buggy input it produces obviously blown-out frames, which surfaces the upstream mistake instead of quietly guessing. It also matches what `inspect-robots-agent` already does in `_png.py` (`clip(0, 1) * 255.0`), so float frames render consistently across plugins. A `float_image_range` knob would mostly exist to accommodate malformed input, and we can always add one later without breaking anything if a real 0–255-float task shows up.

Concretely that means:
- `_to_image`: replace the `arr.max() <= 1.0` conditional with unconditional `np.clip(arr, 0.0, 1.0) * 255.0` for float dtypes (keep the empty-array guard).
- Docstring: "Floats are treated as normalized [0, 1] and scaled to [0, 255]; values outside that range are clipped."
- Tests: the `float_255` case changes — `[0.0, 128.0, 255.0]` now clips to `[0, 255, 255]`. Worth asserting that explicitly so the clipping behavior is pinned down.

Everything else looks good — thanks for the quick turnaround on the earlier feedback!

Amresh-01:
Hi @jeqcho All done — `_to_image` now clips and scales unconditionally, with updated tests. Ready for another look.

jeqcho:
CI ran after workflow approval and it's red, but the cause is mechanical: while this PR was waiting, #42 landed on main with cfg-wiring tests using the same names as yours, and the merge of main kept both copies. Ruff now fails with 4 F811 redefinitions in `plugins/inspect-robots-isaacsim/tests/test_isaacsim_embodiment.py` (`test_ensure_env_wires_cfg_correctly_when_headless` at lines 457/555 and `test_ensure_env_skips_debug_vis_disable_when_not_headless` at 475/571), which is what both failing jobs are.

Fix: drop the duplicated copies, keeping whichever version of each test covers more (they landed via #42, so likely those; if yours assert something extra, fold it in). Your genuinely new tests (`test_ensure_env_wrapped_import_error`, the `_read_success` opt-in cases, the `_to_image` empty/clip cases) stay as is. `uv run ruff check plugins/inspect-robots-isaacsim` locally will confirm before pushing.

Everything else is green and the substance of the PR is done — this is the last hurdle before merge.

Amresh-01:
Hi @jeqcho, I've merged main into the branch, resolved the test duplication, and fixed the line-length violations. Everything is passing and lint-clean locally. Ready for another review/workflow approval!

jeqcho:
Merged! Thank you @Amresh-01 — and thank you especially for how you handled the back-and-forth. This one went through two real rounds of review: the `_read_success` fallback needed a design change rather than a patch, the `_to_image` heuristic turned out to be a genuinely ambiguous problem that needed a convention decision, and at every step you engaged with the substance, asked exactly the right clarifying question (the 0–1 vs range-knob call), and turned the answers around fast. That's the kind of collaboration that makes maintaining this project fun.

What landed is meaningfully better than what any of us had at the start: a safe default for oracle-less Isaac Lab tasks with an explicit opt-in, a float-image convention now consistent across the whole plugin ecosystem, and curated errors instead of raw ImportErrors. The regression tests you wrote pin all of it down.

We'd love to see you around here again — feel free to grab anything in the tracker that catches your eye (https://github.com/robocurve/inspect-robots/issues has a few meaty ones open, and the plugin repos too). Just comment on an issue like you did with #5 and it's yours. 🎉

PR #42, opened 2026-07-10
isaacsim: cover _ensure_env's cfg-wiring contract without Isaac

### LINKED ISSUE #25: isaacsim: _ensure_env has no CI coverage — a cfg-wiring regression would only fail live
Follow-up from the #15 review (non-blocking finding by an independent review pass).

## Problem

The plugin's CI tests exercise the adapter by injecting a fake env directly, so `_ensure_env` itself — the code path #15 fixed — never runs in CI. A future regression that re-drops the `cfg=` argument to `gym.make`, misnames the `parse_env_cfg` import, or breaks `_disable_debug_vis`/`_request_named_obs_terms` wiring would pass CI and only fail on a live Isaac Lab box. That is exactly the failure mode #15 existed to fix: env creation was unconditionally broken and only a live GPU run caught it.

## Proposed fix

Add a no-Isaac unit test that stubs the lazy imports via `sys.modules` (`isaaclab.app`, `gymnasium`, `isaaclab_tasks`, `isaaclab_tasks.utils`) and asserts on `_ensure_env`'s wiring:

- `parse_env_cfg` is called with `(task_id, device=..., num_envs=1)`
- `gym.make` receives that exact cfg object as `cfg=` plus `render_mode="rgb_array"`
- `headless=True` routes the cfg through `_disable_debug_vis`; `headless=False` does not
- the obs group's `concatenate_terms` is set to `False` before `gym.make`

This keeps the plugin job's no-Isaac property (stubs only, no GPU) while pinning the exact contract with Isaac Lab that only live runs verify today.

Tushar2604:
## Summary

CI tests inject a fake env directly for step()/reset() translation, so `_ensure_env`'s own body never actually ran: `parse_env_cfg`'s args, `gym.make` receiving `cfg=`, the `headless` → `_disable_debug_vis` gate, and the named-obs-terms request. A regression here (e.g. #15's missing `cfg=`) would only have failed live.

## Changes

- Stub `gymnasium`/`isaaclab_tasks`(`.utils`) via `sys.modules` and assert the exact contract: `parse_env_cfg(task_id, device=..., num_envs=1)`, `gym.make(task_id, cfg=..., render_mode="rgb_array")`, `debug_vis` flipped only when `headless=True`, `concatenate_terms` always requested regardless of `headless`
- Confirmed `_ensure_env()` is memoized (a second call returns the cached env without re-wiring)
- `CHANGELOG.md` entry under "Unreleased" (there wasn't one yet — main had just cut 0.6.0)

## Checklist

- [x] Tests added/updated (and the `CubePick` mock exercises the change where applicable) — N/A `CubePick`, this is the isaacsim plugin; added `sys.modules` stubs instead, same pattern as the existing `_disable_debug_vis`/`_request_named_obs_terms` tests
- [x] Coverage stays at **100%** (`pytest --cov`) — core untouched, still 100%; this PR only adds plugin tests (plugin has no 100% gate, per CLAUDE.md)
- [x] `mypy` passes (strict) — on `src/inspect_robots_isaacsim` (the only mypy scope CI/pre-commit enforce for this plugin)
- [x] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test — N/A, no public API change, test-only
- [x] Core stays NumPy-only (new deps are optional extras, lazily imported) — unaffected, plugin-test-only change

## Related

Closes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project are documented here. The format is based on
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html). While the version is
 `0.x`, breaking changes may occur on any minor release.
 
+## [Unreleased]
+
+### Added
+
+- isaacsim plugin: `_ensure_env`'s cfg-wiring contract (`parse_env_cfg`'s
+  args, `gym.make(cfg=...)`, the `headless` → `_disable_debug_vis` gate, and
+  the named-obs-terms request) is now exercised in CI via stubbed
+  `gymnasium`/`isaaclab_tasks` modules. Previously only the fake-env-injected
+  `step()`/`reset()` translation was covered, so a regression in `_ensure_env`
+  itself (e.g. #15's missing `cfg=`) would only have failed live (#25).
+
 ## [0.6.0] - 2026-07-10
 
 ### Added
--- a/plugins/inspect-robots-isaacsim/tests/test_isaacsim_embodiment.py
+++ b/plugins/inspect-robots-isaacsim/tests/test_isaacsim_embodiment.py
@@ -308,3 +308,91 @@ def test_request_named_obs_terms_flips_concatenate_flag(
         _request_named_obs_terms(object(), "policy")
     assert len(caplog.records) == 2
     assert all("named observation terms" in r.message for r in caplog.records)
+
+
+# --------------------------------------------------------------------------- #
+# _ensure_env: the cfg-wiring contract, exercised without Isaac (#25).
+#
+# CI injects a fake env directly (_inject_fake) for step()/reset() translation
+# tests, so _ensure_env's own body — parse_env_cfg args, gym.make(cfg=...),
+# the headless->_disable_debug_vis gate, and the named-obs-terms request —
+# never actually ran in CI. A regression here (e.g. dropping cfg=, as #15
+# fixed) would only have failed live. Stub gymnasium/isaaclab_tasks via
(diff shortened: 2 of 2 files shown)


jeqcho:
Thanks, this is a useful test. I mutation-tested it: dropping `cfg=` from `gym.make` (the exact #15 regression) and dropping the `headless` gate both fail your new tests, which is what #25 asked for.

Two assertions are softer than they look though.

1. `fake_make` returns the literal `"fake-env"`, and Python interns that string, so `assert emb._ensure_env() is env` passes even with memoization deleted entirely (I verified). Return `object()` instead, or assert a call count on `calls`.
2. The cfg is recorded by reference and checked after `_ensure_env` returns, so moving `_disable_debug_vis` / `_request_named_obs_terms` to after `gym.make` still passes. Live, that reorder is a silent no-op since Isaac consumes the cfg during `make`. Snapshot `cfg.debug_vis` and `cfg.observations.policy.concatenate_terms` inside `fake_make` and assert on the snapshots.

Both are a few lines each. With those in, this is good to merge.

aris-zhu:
Pushed the two review tweaks directly (maintainer edit) to get this in: gym.make's fake now returns a fresh sentinel per call and snapshots debug_vis / concatenate_terms at make-time, plus a merge with main for the CHANGELOG. Re-ran the mutation checks: deleting the memoization and moving the wiring after gym.make both fail the suite now. Merging once CI is green — thanks for the contribution!

PR #43, opened 2026-07-11
feat: persist LLM agent conversation transcripts and add policy trial…

### LINKED ISSUE #40: agent plugin: persist the LLM conversation transcript per trial
After the first live agent run on hardware (yam_arms, 2026-07-10), the summary `EvalLog` answered "what happened" (steps, duration, status) but not "what did the model think": the agent's conversation — tool calls, tool errors fed back, the model's text between motions — lives only in `LLMAgentPolicy._messages` and is discarded at trial end.

Rollout transcript events (inference/approval/step) reach sinks, but the conversation itself is never persisted anywhere. The Rerun viewer covers live observation; nothing covers post-hoc analysis ("why did it give_up?", "which tool errors did it hit?").

Proposal: the policy writes the final conversation (minus image payloads, which are already in the frame store) to a JSONL file per trial under the log dir, and the path is recorded in trial metadata. Alternatively a `conversation` field on a policy-provided artifact hook in core, if other policies would benefit.

Found while reviewing `logs/adhoc_4d8650ca.json` from the first hardware run.

Adityakk9031:
Resolves #40 

## Overview
This PR addresses the issue where the `LLMAgentPolicy`'s internal reasoning, conversation state, and tool errors were discarded at the end of a trial without being persisted to the log directory, making post-hoc analysis difficult. 

To solve this, a new lifecycle hook has been added to the `Policy` interface, allowing any policy to dump artifacts and append custom metadata to the trial's final record. 

## Changes Made
**1. Core Framework:**
- **Trial Metadata:** Added an extensible `metadata: dict[str, Any]` field to `TrialRecord` (`rollout.py`).
- **EvalLog Persistence:** Added `trial_metadata: tuple[dict[str, Any], ...]` to `SceneResult` (`log.py`) so metadata is properly serialized to and read from the JSON logs.
- **Lifecycle Hook:** Added an optional `on_trial_end(self, record: TrialRecord, log_dir: str)` hook to `PolicyBase` (`policy.py`), which is now invoked by the orchestration loop in `eval.py` right before a trial finishes.

**2. `inspect-robots-agent` Plugin:**
- Implemented the `on_trial_end` hook in `LLMAgentPolicy`.
- When a trial completes, it creates a `transcripts/` folder within the evaluation `log_dir` and writes out the trial's full conversation to a `.jsonl` file (e.g., `transcripts/<scene_id>-e<epoch>.jsonl`).
- Base64 `image_url` payloads are stripped out of the JSONL to prevent massive file bloat (since they are already captured in the frame store).
- The path to the transcript is injected into `record.metadata["transcript"]`, appearing cleanly in the summary `EvalLog`.

## Testing
- Verified that running an eval with the agent policy correctly outputs the `.jsonl` transcript inside the log directory.
- Confirmed the transcript contains the system prompt, tool errors, and model reasoning, but strips the inline image payloads.
- Verified that `EvalLog.json` correctly saves the `trial_metadata` containing the transcript path.


jeqcho:
Thanks, this is a nice minimal take on #40. Routing it through a generic `on_trial_end` hook that mirrors the existing `bind()` duck-typing fits the codebase well, and firing it before `bus.on_trial_end` so sinks see the metadata is the right ordering.

A few things before it can land:

1. CI fails as submitted. `TrialRecord` is unimported in `src/inspect_robots/policy.py:82` (needs a `TYPE_CHECKING` import; a runtime one would be circular) and `Any` is unimported in `eval.py`. `ruff format --check` also flags both touched files.
2. Transcript filenames need a per-run component. Two runs into the same `log_dir` currently overwrite each other's transcripts while the first run's EvalLog still points at the overwritten files. See the run-stamp comment above the `FrameStore` setup in `eval.py`, which exists for this exact problem.
3. Wrap the hook call so a transcript-write failure (unwritable dir, disk full) can't crash the eval and lose the log. The reducer path at eval.py:372 shows the intended pattern.
4. Tests: hook invocation and `trial_metadata` round-trip belong in `test_eval_orchestration.py` / `test_eval_log.py`, and transcript content plus image stripping in the plugin's e2e tests. A CHANGELOG entry too, please.

Glad to re-review after that.

Adityakk9031:
@jeqcho check this

jeqcho:
Hi @Adityakk9031 thanks for this! We merged some new PRs so we are happy to review this once the conflicts are resolved

Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
This is a strong second pass — all four points from my last review are properly addressed, and CI agrees. The run_stamp reuse is exactly the right fix for the filename collisions, the hook-failure handling follows the reducer precedent I pointed at, and I verified the parallel arrays stay aligned on every path (success, errored, cancelled, halting faults) with the hook correctly firing before sinks see the record. The orchestration tests, including the already-errored branch, are careful work.

One thing changed under you through no fault of yours: #101 merged while this was in flight, so the image-free conversation now already lands in the EvalLog via `transcript()`, and `inspect --transcript` renders it. That doesn't sink this PR — the `on_trial_end` hook and `trial_metadata` plumbing are generic artifact-persistence infrastructure worth having regardless — but it does change what the plugin half should look like. Which brings me to the asks:

1. **Reuse `transcript()` for the JSONL** instead of the hand-rolled filter. Right now the file has a second sanitizer with different behavior: `_sanitize()` replaces images with an "[image omitted]" placeholder, while `on_trial_end` drops the parts entirely — so an image-only tool result becomes an empty content list with no trace an image existed, and the side file disagrees with the in-log transcript for the same trial. `for msg in self.transcript() or []` deletes the duplicate and makes the two records identical by construction. With that, the side files earn their keep post-#101 as the grep/stream-friendly form of the same data.

2. **Un-break the frozen-ness test.** In `test_eval_log.py`, the new `trial_metadata.append` line was added inside the existing `pytest.raises(AttributeError)` block, above the old `termination_reasons.append` — which is now unreachable, since the first line raises. Give the new check its own `with` block.

3. **Move the `trial_metadatas.append` calls after the hook runs.** Appending first only works because the list holds a reference to the same dict the hook then mutates. Your test passes today, but the first person to add a defensive copy or eager serialization there loses metadata silently. Append after the hook (or at least leave a comment naming the aliasing).

4. **Docstring the hook's actual contract** on `PolicyBase.on_trial_end`: called for errored and cancelled trials too, before sinks see the record, mutations to `record.metadata` land in the log, and exceptions degrade the run to `status="error"` rather than crashing. That's what an implementer needs to know, per the repo's docstring rule.

Nits, take or leave: wrap the CHANGELOG entries to match the file's line width, drop the stray blank-line changes at the two section headers, and cite (#40) in the entries.

All small and mechanical — happy to approve on the next push. Thanks for sticking with this one through two rounds and a moving main.

Adityakk9031:
@jeqcho hey please review it

jeqcho (approved):
All earlier review points addressed, CI green after the branch update. Merging.

jeqcho:
Merged, thanks Aditya, and thanks for sticking with this across three rounds of review.

Two things in the final version I want to call out. The extra orchestration test that exists purely to cover the already-errored-trial branch shows you understood why the 100% branch-coverage gate is there, and the comment explaining it will save the next contributor a puzzled minute. And hoisting `run_stamp` so frames and transcripts share one identifier was a tidier fix than what I literally asked for; the test pinning `seen_ids[0] == seen_ids[1]` locks that in properly.

One crumb left if you want it: the hook's error note is `policy.on_trial_end failed: {exc}`, which drops the exception class (an exception with an empty message renders as "failed: "). The sibling paths use `{type(exc).__name__}: {exc}`. Tiny follow-up PR if you feel like it, no pressure.

Good to have transcripts persisted at last. Hope to see you on another issue.

Adityakk9031:
@jeqcho  3 prs are  open have a look

PR #63, opened 2026-07-14
docs: heading for the LLM agent example in the quickstart

aris-zhu:
One-heading change: adds `### Drive the robot with an LLM` above the agent-policy example in the quickstart, so readers scanning the README or its table of contents discover that LLMs are supported as policies on real rigs. The example and surrounding prose are unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,8 @@ and actions straight from the eval pipeline, so you watch exactly what the
 policy sees while the robot moves. CLI flags override any default
 (`--no-rerun`, `--no-store-frames`, `--max-steps 300`, ...).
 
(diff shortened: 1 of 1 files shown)


jeqcho:
Good call on the heading — pushed a small follow-up commit: with a single `###` mid-Quickstart, everything after it (sim, the CLI commands, the Python API example) nested under "Drive the robot with an LLM" until the next `##`. Added sibling subheadings (`### Run in simulation`, `### More CLI commands`, `### Python API`) so each part scopes correctly, and dropped the now-dangling "And" openers on the two paragraphs that got headings.

PR #66, opened 2026-07-14
Fix #53: Reuse embodiment verdicts and eliminate duplicate operator prompts

### LINKED ISSUE #53: Operator is prompted twice when a self-confirming embodiment is paired with the operator scorer
## Observed

On the YAM rig, every ad-hoc run ends with two back-to-back verdict prompts for the same episode:

```
Did the robot succeed? [y/N]: y
did the robot succeed? [y/n/partial/skip] (partial scores as failure) y
```

- Prompt 1 is the **embodiment's** (`inspect_robots_yam/operator.py: confirm_success`), per the documented real-hardware contract: no privileged success oracle, so the adapter asks the operator and converts the answer into `StepResult(terminated=True, termination_reason="success")`.
- Prompt 2 is the **CLI's** `_prompt_operator` (`cli.py:340`), wired whenever the resolved scorer is `operator` (which is also the ad-hoc default). It records `operator_judgement` on the trial record, unaware the embodiment already collected the same verdict.

Both encode the same fact; scorers like KitchenBench's `task_success` (and `success_at_end`) accept either signal.

## Why it matters

- The operator answers the same question twice per episode, forever — on a 50-scene benchmark run that is 100 prompts, half of them noise.
- Worse than annoying: the two answers can disagree (typo, changed mind), and nothing reconciles them — the scorer silently prefers one signal and the log carries both.

## Proposal

`_prompt_operator` (or the `operator` scorer path generally) should skip the prompt when the trial already carries a definitive embodiment-level verdict:

- If `termination_reason == "success"` → record an affirmative `operator_judgement` (source: embodiment) without prompting.
- Only prompt when the episode ended without a verdict (timeout/truncation on an embodiment that does not self-confirm) — which is exactly the case the `operator` scorer exists for.

Alternative/simpler: document that self-confirming embodiments should be paired with `success_at_end` rather than `operator` (that combination already produces a single prompt today), and have the CLI warn on the redundant pairing. The skip-when-verdict-exists behavior seems strictly better, though: `operator` stays a safe default for any embodiment.

Workaround meanwhile: `inspect-robots config set scorer success_at_end` on rigs whose embodiment self-confirms.

Related context: robocurve/inspect-robots#50 (first-run field report; this is the same "each layer independently talks to the operator" theme).

theonlypal:
## Description

This PR fixes #53 by making an existing embodiment-level terminal outcome the single source of truth for ad-hoc runs that use the `operator` scorer.

Self-confirming embodiments such as YAM already ask the operator for a verdict and return a terminated `StepResult`. The CLI previously ignored that outcome and asked the same question again during `before_scoring`, creating redundant prompts and allowing contradictory answers in the same log.

The CLI now reuses a definitive embodiment outcome before considering terminal input. This applies with an interactive TTY, non-interactive stdin, and `--no-prompt`. A terminal prompt is used only when the rollout ended without a definitive embodiment verdict.

## Changes

- Reuse a terminated embodiment outcome as the recorded `operator_judgement`.
- Map `termination_reason == "success"` to `yes` and explicit terminal failure reasons to `no`.
- Record the adopted verdict as an operator transcript event with `source: embodiment`.
- Preserve embodiment verdicts when stdin is non-interactive or `--no-prompt` is set.
- Keep terminal prompting for timeout, truncation, or a terminated result without a reason.
- Preserve existing typo retry, `partial`, `skip`, and EOF behavior.
- Document the single-verdict behavior and add an Unreleased changelog entry.
- Add unit and end-to-end regression tests covering prompted and non-prompted paths.

## Verification

- Ruff lint and formatting pass.
- Strict mypy passes.
- 409 tests pass on Python 3.11 and Python 3.12.
- Line and branch coverage remain at 100%.
- 409 tests pass on Python 3.10 with NumPy 1.x and Python 3.13 with NumPy 2.x.
- All 94 first-party plugin tests pass.
- Real `rerun-sdk` compatibility tests pass.
- Pre-commit and pre-push hooks pass.
- Strict MkDocs build passes.
- Core-only NumPy import hygiene passes.
- `uv lock --check` passes.
- Wheel and source-distribution builds pass.

Closes #53


jeqcho:
Thank you so much for this, and I owe you an apology on timing: I landed my own fix for #53 in #69 the same day you opened this, so the duplicate-prompt bug is already fixed on main. That's why this branch shows conflicts now.

I want to be clear this is no reflection on your work — you read the problem exactly right. Making the embodiment's verdict the single source of truth is precisely the shape the merged fix took, and details like recording the adopted verdict as a transcript event with `source: embodiment` were genuinely nice touches. The whole PR was clean, well-tested, and green on CI. It just got raced.

Closing as superseded by #69, but please keep the contributions coming — this one was good enough that I'd be happy to see you pick up any open issue that interests you. And next time, feel free to ping me on the issue before you start so we don't double up. Thanks again!

PR #67, opened 2026-07-14
fix(rollout): formalize self-paced-only control-rate contract

### LINKED ISSUE #3: rollout: control-rate pacing is dead code — R1 rate precedence unimplemented
## Problem

`rollout()` accepts `control_hz` (wired from `task.control_hz`) but never paces: `src/inspect_robots/rollout.py:104-117` computes `_ = _effective_control_hz(...)` and `_ = SELF_PACED` as coverage-satisfying no-ops, and the loop at `rollout.py:186-189` explicitly does not sleep. `chunk.control_hz` is never consulted, despite plan 0001 R1's rate precedence being "chunk → task → embodiment".

Docs (`docs/guide/concepts.md`) tell adapter authors the framework paces; on a real robot it does not. Both hardware adapters (inspect-robots-so101, inspect-robots-yam) noticed and work around it with `SELF_PACED` + internal sleeps — which happens to be correct today, but only by reading the source.

## Decision needed

Either implement pacing in the rollout loop honoring the R1 precedence, or codify SELF_PACED-everywhere and update R1 + docs to match. The current state (accepted-but-ignored knob) misleads adapter authors and eval reproducibility records.

Tushar2604:
## Summary

`rollout()` accepted `control_hz` (from `task.control_hz`) but never paced the
loop: `_effective_control_hz()` computed R1's chunk → task → embodiment rate
precedence, but nothing called it and the loop never slept. Docs and
docstrings claimed "the framework owns pacing," which was false on real
hardware — both hardware adapters (so101, yam) already self-pace internally
as a workaround. Formalize that as the actual contract instead of
implementing real pacing, which would need its own design pass to keep the
CubePick-driven test suite fast (CubePick declares `control_hz=10.0`; real
pacing would add ~0.1s/step to every test).

## Changes

- Remove the dead `_effective_control_hz` helper and `rollout()`'s unused
  `control_hz` param.
- Remove `Task.control_hz` (its only reader was the now-removed rollout
  param) — public API break.
- Rewrite the "framework owns pacing" docstrings (`embodiment.py`,
  `docs/guide/policies-and-embodiments.md`) from the adapter author's point
  of view: rollout paces nothing; an embodiment needing real-time cadence
  paces itself in `step()` and declares `"self_paced"` to document that.
- Update plan 0001 §9 R1 in place, keeping the original resolution visible
  under a dated inline reversal note with rationale.
- `compat`'s policy/embodiment rate-mismatch warning is untouched.

## Checklist

- [x] `pre-commit install` done; hooks pass (or ran `pre-commit run --all-files`) — `pre-commit run --all-files` fails in this environment because its hooks shell out to `uv`, which isn't installed here; ran `ruff check .`, `ruff format --check .`, and `mypy` directly instead (all pass)
- [x] Tests added/updated (and the `CubePick` mock exercises the change where applicable) — removed the now-meaningless `test_effective_control_hz_precedence`; no new behavior to cover since this only removes dead code/params and rewrites docs
- [x] Coverage stays at **100%** (`pytest --cov`) — 100% on every touched module (`rollout.py`, `task.py`, `eval.py`, `embodiment.py`); full local run shows 99.03% only because `test_setup.py`'s symlink tests skip without Windows dev-mode privilege (confirmed pre-existing on `main`, unrelated — CI's blocking gate runs Linux/macOS)
- [x] `mypy` passes (strict)
- [x] `CHANGELOG.md` updated under "Unreleased" — added a `Removed` section (none existed yet)
- [x] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test — N/A to `__all__` (`Task` stays exported; only a field was removed, which the snapshot test doesn't check at field level); flagged as breaking in the changelog instead, per @jeqcho's request the next release should be cut as **minor**
- [x] Core stays NumPy-only (new deps are optional extras, lazily imported) — unaffected, no new deps

## Related

Direction confirmed with @jeqcho in the issue thread before implementation.

Closes #3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ All notable changes to this project are documented here. The format is based on
 
 ## [Unreleased]
 
+### Removed
+
+- **`Task.control_hz`** (breaking). `rollout()` never actually paced the
+  control loop to it — `_effective_control_hz` was dead code and the loop
+  never slept — so the field only misled adapter authors and eval
+  reproducibility records. The rollout now documents plainly that it applies
+  no wall-clock pacing of its own; an embodiment that needs real-time cadence
+  paces itself in `step()` and declares the `"self_paced"` capability. Plan
+  0001 §9 R1 is updated to match (see the inline reversal note there for
+  rationale). `compat`'s policy/embodiment rate-mismatch warning is
+  unaffected.
+
 ### Added
 
 - Plugin-declared embodiment device slots for V4L2 cameras, SocketCAN
--- a/docs/guide/policies-and-embodiments.md
+++ b/docs/guide/policies-and-embodiments.md
@@ -80,8 +80,8 @@ class MyArm:
         ...
 
     def step(self, action: Action) -> StepResult:
-        # Returns as soon as the command is issued; the framework paces the loop
-        # unless this embodiment declares the "self_paced" capability.
+        # The rollout applies no wall-clock pacing of its own. If this embodiment
+        # needs real-time cadence, pace it here and declare "self_paced".
         ...
 
     def close(self) -> None:
--- a/plans/0001-foundation-design.md
+++ b/plans/0001-foundation-design.md
@@ -475,13 +475,23 @@ issues clustered in the rollout↔controller↔record triad. These resolutions a
 **binding** and supersede the prose above wherever they conflict:
 
 - **R1 — `step()` timing contract.** `Embodiment.step()` **returns as soon as the
-  command is issued** (it does *not* block until the control period elapses). The
-  **framework owns pacing** to the *effective rate* = `chunk.control_hz` →
-  `task.control_hz` → `embodiment.info.control_hz` (first non-None). A real-robot
-  adapter whose `step()` inherently blocks at the hardware rate declares
-  capability `"self_paced"`, and the rollout then skips its own pacing. `compat`
-  warns when the policy's emitted rate and the embodiment's native rate differ by
-  more than a tolerance.
+  command is issued** (it does *not* block until the control period elapses).
+  **Reversed 2026-07-14** (see issue tracker: rollout pacing was accepted-but-ignored
+  in practice — `_effective_control_hz`/`Task.control_hz` were dead code, and both
+  hardware adapters already self-paced as a workaround). The rollout applies **no
+  wall-clock pacing of its own**; an embodiment that needs real-time cadence paces
+  itself inside `step()` and declares capability `"self_paced"` to document that it
+  does. `Task.control_hz` is removed — the framework never had a rate of its own to
+  resolve. `compat` still warns when the policy's emitted rate and the embodiment's
+  native rate differ by more than a tolerance (unaffected by this reversal).
+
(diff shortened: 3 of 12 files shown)


jeqcho (changes requested):
Thanks for this — the direction is exactly what we agreed on the issue, and the execution is careful. I checked the blast radius myself before reviewing: nothing constructs `Task(control_hz=...)` in this repo, any plugin, or the downstream repos (kitchenbench, so101, yam), and `EvalLog` never serialized the field, so old logs and `inspect-robots video` are unaffected. The plan 0001 edit is the right way to touch a binding resolution, and I agree with cutting the next release as minor.

One real blocker, two nits:

**Blocker: the compat warning still promises pacing.** `src/inspect_robots/compat.py:163` emits, at runtime, "framework will pace to the effective rate" — which is precisely the false claim this PR exists to remove, except this one is user-facing rather than a docstring. The comment above it (`compat.py:155`, "only warn, since the framework paces") has the same stale rationale. Keeping the warning is right; the message needs rewording to something like "neither rate is enforced; reconcile in the policy or embodiment". `tests/test_coverage_completion.py:433` exercises this path, so check whether it pins the message text.

**Nit:** two leftover "framework paces" claims in the isaacsim plugin: `plugins/inspect-robots-isaacsim/src/inspect_robots_isaacsim/embodiment.py:53` ("Inspect Robots owns wall-clock pacing") and `tests/test_isaacsim_embodiment.py:65` ("# sim is framework-paced"). Comments only, but they contradict the new contract in the exact way the PR is trying to stamp out.

**Take-or-leave:** the `ActionChunk.control_hz` docstring (`src/inspect_robots/types.py:65`) says `None` "defers to the embodiment's native rate", which reads like residue of the removed precedence chain. Defensible as-is since the agent plugin has its own fallback, so your call.

Also: the branch is behind main and the ruleset requires up to date, so it'll need a rebase or update-branch before merge anyway — good moment to fold the fixes in. Happy to approve once the compat message is fixed.

Tushar2604:
Thanks for the detailed review! I've addressed the blocker, updated the stale comments, and rebased the branch with the latest `main`. The changes are now ready for another look.

jeqcho (approved):
This is exactly what I hoped the second pass would look like. The compat message now tells users the truth, the isaacsim comment got *better* than a straight deletion (spelling out that sims run at full speed is genuinely useful to the next reader), and you caught the stale TaskEnvelope rationale on your own sweep, which I'd missed in my review. That kind of follow-through is rare and appreciated.

Merging now. The next release goes out as minor per the changelog note.

Thank you for taking on an issue that was mostly about having the discipline to delete things and tell the truth in docs. Those PRs don't win glory but they keep a codebase honest. I'd be glad to see you pick up anything else in the tracker that catches your eye, and feel free to ping me on an issue before starting so I can confirm direction like we did here. It clearly paid off.

Tushar2604:
Thanks! Really appreciate all the feedback and guidance throughout the review. I'll definitely keep reaching out on issues beforehand—it made the whole process much smoother. Looking forward to contributing more!

PR #68, opened 2026-07-14
feat(cli): add eval-set for running multiple tasks in one invocation

### LINKED ISSUE #45: CLI: run a set of tasks in one invocation (eval-set / multi-task run)
## Feature request

There is currently no way to run an entire benchmark from the CLI. `inspect-robots run` accepts exactly one `--task` (no repeatable flag, no glob), so running all 10 KitchenBench tasks means a shell loop or dropping into Python:

```python
from inspect_robots import eval_set
from kitchenbench import TASK_FACTORIES

ok, logs = eval_set([f"kitchenbench/{k}" for k in sorted(TASK_FACTORIES)], "kitchen_scripted", "kitchen")
```

The Python API already has `eval_set()` (mirrors Inspect AI, returns `(success, logs)`), but the CLI can't drive it.

## Prior art

Inspect AI makes the benchmark the unit of invocation:

```console
$ inspect eval inspect_evals/browse_comp --model openai/gpt-5
```

A benchmark-suite feel is the point of the WorldEvals collection; "run KitchenBench" should be one command.

## Proposed UX (pick any or several)

1. **Repeatable flag**: `inspect-robots run --task kitchenbench/pour_pasta --task kitchenbench/stack ...` — minimal change, verbose for 10 tasks.
2. **Glob/prefix**: `inspect-robots run --task 'kitchenbench/*'` — matches against the registry (entry-point discovery already namespaces tasks as `<benchmark>/<key>`, so a benchmark IS a prefix). Today this errors with `no task named 'kitchenbench/*'`.
3. **A dedicated `eval-set` subcommand**: thin CLI wrapper over the existing `eval_set()`, including its `retry_attempts`/`fail_on_error` semantics and a set-level summary (per-task `task_success` table, one exit code).

Option 2 + 3 together would give the Inspect-Evals-style one-liner:

```console
$ inspect-robots eval-set 'kitchenbench/*' --policy xpolicylab -P url=ws://host:19000 --embodiment yam_arms
```

## Notes

- Log handling: `eval_set` writes one `EvalLog` per task; the CLI should print the log dir once and a compact per-task results table rather than 10 full summaries.
- Guardrails/config defaults should apply exactly as they do for single-task `run`.
- KitchenBench's README currently documents the Python `eval_set` snippet as the way to run the full benchmark; once this lands it can switch to the one-liner.

Tushar2604:
## Summary

There was no way to run a whole benchmark from the CLI: `run` accepts exactly
one `--task`, so running e.g. all 10 KitchenBench tasks meant a shell loop or
dropping into Python to call `eval_set()` directly, even though the Python
API already has it.

## Changes

- Add `inspect-robots eval-set TASK [TASK ...]`: task names are matched
  exactly or by `fnmatch` glob (entry-point discovery already namespaces
  tasks as `<benchmark>/<key>`, so a benchmark name is a ready-made prefix,
  e.g. `'kitchenbench/*'`), deduplicated across overlapping patterns, and
  driven through `eval_set()` against one resolved policy/embodiment pair.
- Resolve the embodiment **once** for the whole set rather than once per task
  — `eval_set()` reconnects a string-embodiment on every call, which would
  mean reconnecting a real robot between every task in the set.
- Output: one status line, a compact `[status] task  metrics-or-error` row
  per task, and the shared log dir printed once — not N full run summaries.
- `-P`/`-E`/`--sim`/`--epochs`/`--fail-on-error`/`--store-frames`/
  `--disable-guardrails`/`--max-action-delta` behave exactly as they do for
  `run`, applied to every matched task.
- `docs/guide/cli.md` gets a new `eval-set` section; `CHANGELOG.md` gets an
  `Added` entry.

**Deliberately out of scope for this pass** (called out in the issue and in
code comments, not silently dropped):
- `--rerun`: `eval_set()` doesn't expose a `sinks` parameter to plug a live
  viewer into, and streaming several back-to-back tasks into one viewer
  window is a separate design question from running the set at all.
- `-T` (per-task constructor args): the issue frames this as a "thin CLI
  wrapper," and per-task args need their own disambiguation design.

## Checklist

- [x] `pre-commit install` done; hooks pass (or ran `pre-commit run --all-files`) — `pre-commit run --all-files` fails in this environment because its hooks shell out to `uv`, which isn't installed here; ran `ruff check .`, `ruff format --check .`, and `mypy` directly instead (all pass)
- [x] Tests added/updated (and the `CubePick` mock exercises the change where applicable) — 21 new tests in `test_registry_cli.py`: exact/glob/dedup matching, zero-match error, epochs override, `--sim`, guardrail conflict/disable/degrade branches, aggregate-error-on-one-task-failing, and the PolicyError-without-`--fail-on-error` case where a task row has neither a metric nor an error to show
- [x] Coverage stays at **100%** (`pytest --cov`) — `cli.py` is 100%; full local run shows 99.07% only because `test_setup.py`'s symlink tests skip without Windows dev-mode privilege (confirmed pre-existing on `main`, unrelated — CI's blocking gate runs Linux/macOS)
- [x] `mypy` passes (strict)
- [x] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test — N/A, CLI-only, no change to the `inspect_robots` package's public symbols
- [x] Core stays NumPy-only (new deps are optional extras, lazily imported) — unaffected, no new deps; also ran a strict `mkdocs build --strict` locally to confirm the new docs section doesn't break the doc build

## Related

Closes #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 
+- **`inspect-robots eval-set TASK [TASK ...]`**: run several registered tasks
+  against one resolved policy/embodiment pair in a single invocation, matching
+  task names exactly or by shell-quoted `fnmatch` glob (e.g.
+  `'kitchenbench/*'`). Thin CLI wrapper over
+  [`eval_set`][inspect_robots.eval.eval_set] that resolves the embodiment once
+  for the whole set rather than once per task, and prints one status line plus
+  a compact per-task row instead of a full summary per task (#45).
 - Plugin-declared embodiment device slots for V4L2 cameras, SocketCAN
   interfaces, and serial devices. `inspect-robots setup` probes and interviews
   declared slots, enforces grouped all-or-none assignments, and suggests udev
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -190,6 +190,40 @@ run a single ad-hoc scene.
 
 The exit code is `0` on a successful eval, `1` otherwise.
 
+## `inspect-robots eval-set`
+
+Run several registered tasks against one resolved policy/embodiment pair in a
+single invocation — the CLI counterpart of
+[`eval_set`][inspect_robots.eval.eval_set]. Task names are matched exactly, or
+by shell-quoted `fnmatch` glob (entry-point discovery namespaces tasks as
+`<benchmark>/<key>`, so a benchmark name is a ready-made prefix):
+
+```bash
+inspect-robots eval-set 'kitchenbench/*' --policy xpolicylab -P url=ws://host:19000 \
+             --embodiment yam_arms
+inspect-robots eval-set cubepick-reach my-other-task --policy scripted --embodiment cubepick
+```
+
+Multiple patterns may match the same task; it still runs once. A pattern that
+matches nothing is an error listing every registered task. `--policy` and
+`--embodiment` (and `-P`/`-E`, `--sim`, `--epochs`, `--fail-on-error`,
+`--store-frames`, `--disable-guardrails`, `--max-action-delta`) apply exactly
+as they do for `run`, to every matched task — there is no per-task `-T` in
+this release. The embodiment is resolved once for the whole set, not once per
+task, so a real robot is not reconnected between tasks.
+
+Rather than one full summary per task, the CLI prints the resolved
+policy/embodiment, one status line for the whole set, a compact `[status]
+task_name  metrics-or-error` row per task, and the shared log directory once
+(`eval_set` still writes one `EvalLog` per task inside it). The exit code is
+`0` iff every task's log has `status == "success"`.
+
+`--retry-attempts` is accepted and threaded through to `eval_set()`, whose
+resumption-of-a-partial-run behavior is reserved for a follow-up: passing it
+today does not yet skip already-finished scenes. `--rerun`'s live viewer
+is not offered for `eval-set`: streaming several back-to-back tasks into one
+viewer window is a separate design question from running the set at all.
+
 ## `inspect-robots doctor`
 
 `doctor` reports a registered embodiment's missing declared runtime modules
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -8,6 +8,11 @@ Subcommands:
   components from the registry. Pass constructor args with ``-T/-P/-E k=v``;
   ``--epochs``, ``--fail-on-error``, and ``--store-frames`` tune the run. The
   written log's path is printed at the end.
+- ``inspect-robots eval-set TASK [TASK ...] --policy P --embodiment E`` — run several
+  registered tasks (exact names or ``fnmatch`` globs, e.g. ``'kitchenbench/*'``) against
+  one resolved policy/embodiment pair via
+  [`eval_set`][inspect_robots.eval.eval_set]. Prints one status line and a compact
+  per-task row instead of a full summary per task.
 - ``inspect-robots inspect LOG.json`` — print a saved eval log.
 - ``inspect-robots setup`` — interactively configure defaults and camera devices.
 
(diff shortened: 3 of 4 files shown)


jeqcho:
Hi @Tushar2604 we are happy to review this once the conflicts are resolved!

Tushar2604:
Hello @jeqcho , Thanks for the review , I will resolve the conflicts shortly .

jeqcho (changes requested):
Thanks — the core design here is right, and it's the part that matters: resolving the embodiment once so a real robot isn't reconnected between tasks is the actual feature, and it's implemented correctly. Resolution order is actually slightly better than `run`'s (the embodiment is resolved last, right before the try, so there's no leak window for --epochs/task validation to escape through). I also checked the things that tend to bite: every kwarg you pass exists on `eval_set()` with matching semantics; `EvalLog.results` is non-optional so the summary can't crash on errored tasks; the filename in the hint matches `json_log.py`'s actual convention; and skipping the operator prompt is correct parity, not a gap — `run` only wires it for ad-hoc runs. The scoping honesty on `--rerun`/`-T`/`--retry-attempts` and the test coverage (including the zero-scene and all-trials-errored edges) are appreciated.

What I'd like before merge:

1. **Handle Ctrl-C.** `_cmd_run` catches KeyboardInterrupt, points at the partial log, and returns 130 — behavior #118 added precisely because hardware evals get interrupted. A 10-task benchmark on a real robot is now the longest-running thing this CLI can do, and Ctrl-C mid-set dumps a raw traceback. Your finally still de-energizes the arm, and #118 means partial logs are on disk — the user just never learns where. Mirror `run`'s handler, pointing at the log dir.

2. **Fold the duplication while you rebase.** The branch conflicts with main (`run`'s output moved under you in #125, and `cli.py` grew `video`), and the ~100 duplicated lines are why: nine shared flags and the whole resolution preamble (defaults → pick → config-args → kv merge → resolve → guardrails print) are copy-pasted from `_cmd_run`, so every change to `run` now silently diverges `eval-set`. A `_add_shared_eval_args(parser)` helper plus extracting the resolution block would make the next flag land in both commands for free. Since a rebase is unavoidable anyway, this is the moment.

3. **Match #125's summary conventions on rebase.** Main's `run` now prints a styled run-status label with an outcome line; make `eval-set`'s status line read like the same CLI.

Take-or-leave: the failure hint prints a `<task>_<id>.json` template rather than a real path because `eval_set()` doesn't expose sink paths. Globbing the log dir for the freshest matches (or just printing the failed task's slug) would be a cheap upgrade, but fine to defer.

Happy to approve once 1–3 are in.

Tushar2604:
@jeqcho Done--made the changes you asked for in `aa35700`.

* Mirrored `run`'s Ctrl-C handling so it now points to the log directory, exits with code 130, and still closes the embodiment in the `finally` block. Added a test for it.
* Refactored the shared CLI logic into common helpers so `run` and `eval-set` stay in sync, while keeping embodiment resolution last to avoid any leak window.
* Updated `eval-set` to use the same status output as `run` (`_display_status` with `[completed]`/`[error]`).
* Left the `<task>_<id>.json` hint unchanged for now, as discussed.

Also re-ran everything on Linux: 100% coverage, `mypy --strict`, and `mkdocs --strict` all pass.

jeqcho (approved):
Approved, and thank you. This is the second PR today where your response to review went past patching the lines I named: migrating `run` onto the shared helpers instead of leaving the old copy in place is what actually kills the drift risk, and nobody asked you to do it. I also noticed the leak-window property survived the extraction and got documented as a contract in `_resolve_components`, and the plan-0008 rationale moved into the guardrails helper instead of getting lost. That's the difference between addressing feedback and understanding it.

The Ctrl-C test is well built too — patching `inspect_robots.eval_set` only works because the command imports it lazily, and you clearly knew that.

Merging now. Between this and #67 you've had a great couple of days in this repo; I'd be glad to see you take on anything else in the tracker. eval-set resumption (`--retry-attempts` actually honoring partial runs) is a natural follow-up to your own work here if it interests you.

Tushar2604:
@jeqcho Thanks! I really appreciate the detailed reviews and encouragement. Looking forward to picking up another issue soon, and I'll definitely take a look at the resumption issue.

PR #72, opened 2026-07-14
fix: auto-load .env from the working directory in the CLI

aris-zhu:
## Problem

With `ANTHROPIC_API_KEY` present in `./.env`:

```
$ uv run inspect-robots "place the fork on the plate" --policy agent -P model=anthropic/claude-fable-5
no API key found for model 'anthropic/claude-fable-5'.
```

The agent policy resolves providers from `os.environ` only; nothing in the framework (or `uv run`) loads `.env`, so keys placed there were invisible. Inspect AI, which this framework mirrors, auto-loads `.env` at its entry points, so this is the workflow users expect.

## Fix

- New private `inspect_robots._dotenv`: a dependency-free `.env` parser (`KEY=VALUE`, `#` comments, optional `export` prefix, matching-quote stripping, malformed lines skipped) plus `init_dotenv()`, which applies values from `./.env` without overriding real environment variables. No python-dotenv dep: core stays NumPy-only.
- `cli.py:main()` calls `init_dotenv(os.environ)` before dispatch, so every subcommand (and `load_defaults`, which reads env) sees the values.
- README note and module-map row.

## Verification

- 422 tests pass at 100% coverage; ruff, ruff format, and strict mypy clean.
- End-to-end: with `ANTHROPIC_API_KEY=sk-test-123` in `./.env` and the variable unset in the shell, the same command now gets past provider resolution and reaches api.anthropic.com (the eval log records the expected HTTP 401 for the stub key). Without `.env` it still fails with the guided ConfigError.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,9 @@ motion chunk per call:
 uv run inspect-robots "place the fork on the plate" --policy agent -P model=anthropic/claude-fable-5
 ```
 
+Keys can live in a `.env` file in the working directory: the CLI loads it
+automatically, and real environment variables take precedence over its values.
+
 ### Run in simulation
 
 The same instruction runs on your configured simulator instead of the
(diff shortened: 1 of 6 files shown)


aris-zhu:
An independent review pass (fresh-context agent) found four confirmed issues, all fixed in the follow-up commit:

1. A non-UTF-8 `.env` raised an uncaught `UnicodeDecodeError` at the top of `main()`, crashing every subcommand. Now caught alongside `OSError`, with a test.
2. A UTF-8 BOM (Windows Notepad) left U+FEFF glued to the first key, silently ignoring it. Now decoded with `utf-8-sig`, with a test.
3. Test hermeticity: every in-tree `main([...])` call was reading the developer's real `./.env` into `os.environ`. An autouse `conftest.py` fixture now no-ops dotenv loading in tests; the wiring test opts back in explicitly.
4. The wiring test's sentinel env var leaked past teardown (`delenv(raising=False)` on a missing key records no restore). Fixed with a setenv-then-delenv pair.

Also aligned unquoted-value inline comments (`KEY=sk-123 # prod`) with python-dotenv semantics: the comment is stripped instead of embedded in the value.

All gates re-verified: 425 tests at 100% coverage, ruff, ruff format, strict mypy.

Issue #71, opened 2026-07-14
setup: hint when the V4L2 probe is inconclusive and the camera listing falls back to all entries


aris-zhu:
Follow-up from the PR #70 review (Minor #5).

`_scan_cameras` discards the distinction between probe verdicts: `None` (could not probe: not in the `video` group, non-Linux, odd udev permissions) and `False` (probed fine, not a color-capture node). When no node probes `True`, the wizard silently lists every entry, including metadata nodes, and the "only N by-id entries for M detected cameras" hint computes M from raw entry counts, overcounting in fallback mode.

Suggested improvement: thread the per-entry verdicts up to `_camera_section` so the wizard can print a one-line yellow hint such as "couldn't probe /dev/v4l nodes (not in the `video` group?); listing everything", and base the count hint on probed cameras when verdicts are available. Also consider listing `None` nodes alongside `True` nodes when both occur in one directory (currently a per-node permission oddity hides the unprobeable node).

Low priority: all paths have a manual-entry escape hatch and the fallback errs toward showing more, not less.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Amresh-01:
Hi @aris-zhu, I noticed you opened this and the related PR #83. Are you planning to implement this yourself, or would you welcome someone else taking it? I'd be interested if it's up for grabs.

PR #83, opened 2026-07-14
fix: setup wizard probes Enter-accepted stale camera devices

### LINKED ISSUE #82: Setup wizard Enter-accepts '(current, not detected)' camera devices without probing them
## What happens

When the wizard suggests a camera slot's current config value, pressing Enter accepts it even when the value is flagged `(current, not detected)`. The accepted path is never run through the color-capture probe added in #70, so a stale or wrong device (for example a RealSense depth node from a pre-#70 config) sails through setup and only fails minutes later at runtime as `EmbodimentFault: cannot open top_cam`.

Related: the wizard's device listing defaults to `/dev/v4l/by-id`, but udev's by-id table can be incomplete or index-scrambled after re-enumeration. On a rig today, a D435's color node (`/dev/video16`) had no by-id symlink at all; the correct choice only existed under by-path. A user who does not know to press `p` hits a dead end: the correct device is not selectable from the default listing, while the wrong one is Enter-acceptable. This is the sharp end of the inconclusive-hints problem in #71.

## Proposed fix

- On Enter-accept of a not-detected current value, probe it with the V4L2 color-capture check. If it fails, print a red warning stating why (not color-capable, or missing) and re-prompt once. A second Enter keeps the value, so headless re-runs and unusual devices are not blocked.
- When capture-capable nodes exist under by-path but have no by-id entry, print a per-slot notice naming the affected nodes and suggesting `p`, instead of relying on the user to discover the toggle.

aris-zhu:
Closes #82. Related: #71, #70.

Spec: plans/0013-setup-probe-accepted-current.md (two adversarial critique
rounds; round 2 verdict ready).

- Enter-accepting a v4l2 current value absent from both scans now validates
  it: a missing path or a definitive probe-False (color-capture check from
  #70) prints one yellow warning and re-prompts; a second Enter always
  accepts. Probe True or inconclusive (None) accepts silently, so Windows
  workstations, foreign-machine configs, and unusual devices see zero new
  friction; same for values present in either listing.
- The by-id/by-path count-mismatch hint is now a shared helper gated on
  by-path entries whose resolved targets have no by-id entry. It names the
  affected nodes (the motivating rig needed to find /dev/video16), states
  both possible causes (udev serial collision or missing symlink), and
  drops the 'p' instruction when the by-path listing is already active.
  The prompt's own 'p' advertisement (advertise_path_toggle) is unchanged.

Deviation from the issue text: the warning is yellow, not red; the wizard
palette defines no red and yellow is its warning convention throughout.

Gates: ruff check, ruff format --check, mypy strict, pytest --cov all green;
459 passed (up from 451), coverage 100%.

### DIFF
--- a/plans/0013-setup-probe-accepted-current.md
+++ b/plans/0013-setup-probe-accepted-current.md
@@ -1,7 +1,9 @@
 # Probe Enter-accepted stale camera devices in the setup wizard
 
 Date: 2026-07-14
-Status: draft; subagent critique round 1 in progress
+Status: revised after subagent critique round 1 (probe-True acceptance, both
+listings checked before probing, discriminator dropped, dual hint sites,
+explicit test updates); round 2 pending
 
 ## Problem
 
@@ -15,9 +17,9 @@ rig on 2026-07-14; see #82).
 Separately, when udev's by-id table is missing capture-capable nodes that
 by-path has (same rig: the D435 color node `/dev/video16` had no by-id
 symlink at all), the only clue is the count-mismatch hint, whose stated cause
-("identical cameras without serials collide") is wrong for this failure
-mode. The correct device is not selectable from the default listing while
-the wrong one is Enter-acceptable. This is the sharp end of #71.
+("identical cameras without serials collide") is one possible cause of
+several, and whose remedy (the `p` toggle) is easy to miss. This is the
+sharp end of #71.
 
 ## Requirements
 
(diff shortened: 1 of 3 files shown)


aris-zhu:
Fresh-eye review round 1: blocking finding (vacuous hint-silence assertion) fixed in the follow-up commit and verified by re-running the reviewer's mutation (removing the hint's early return now fails the test). Non-blocking notes recorded for future work, none regressions: a declined duplicate-confirm can re-probe a True/None node on the next Enter; a dangling by-id symlink draws the foreign-machine wording rather than an unplugged-camera hint; the hint's name list is one unbounded line on many-camera rigs. First two are natural follow-ons for #71.

PR #144, opened 2026-07-17
fix(approver): allow delta pose modes with clampable rotation deltas

### LINKED ISSUE #143: DeltaLimitApprover rejects delta pose modes with clampable rotation deltas
DeltaLimitApprover raises ValueError for any pose mode whose rotation_repr is not in {none, rot6d} (approver.py: _POSE_MODES includes eef_delta_pose; the check fires before bounds or max_delta are consulted). For ABSOLUTE pose modes that is sound: per-dim clamping of absolute euler orientations has wraparound and axis-coupling problems. But for DISPLACEMENT modes (eef_delta_pose, eef_delta_pos) the per-dim quantities are small rotation deltas, and clamping them per-dim is well defined.

Consequences today, hit while designing inspect-robots-widowx (BridgeData V2 speaks 7-D xyz+euler deltas):
- conformance's guardrail probe turns the ValueError into a doctor error, so any euler-delta embodiment is reported non-conformant.
- cli catches the ValueError and silently degrades every CLI run to clamp-only; --max-action-delta cannot restore delta limiting.

Proposal: restrict the rotation-repr rejection to absolute pose modes; for displacement modes, treat rotation dims like any other bounded displacement dim.

Workaround adopted by inspect-robots-widowx v1 (yam precedent): declare rotation_repr="none" with droll/dpitch/dyaw dim_labels. Once this lands in a release, widowx can declare euler_xyz honestly and bump its pin.

Tushar2604:
## Problem

`DeltaLimitApprover` raised `ValueError` for any pose mode whose `rotation_repr` wasn't in `{none, rot6d}`. The check fired for `_POSE_MODES`, which includes the displacement mode `eef_delta_pose`.

For absolute orientations that's sound, since per-dimension clamping of absolute Euler/quaternion representations has wraparound and axis-coupling problems. However, a displacement pose mode's per-dimension quantities are small rotation deltas, which clamp per dimension just fine.

The false rejection had two downstream effects:

* Conformance reported Euler-delta embodiments as non-conformant (e.g. `inspect-robots-widowx` / BridgeData V2's 7-D `xyz + euler` deltas).
* CLI runs silently degraded to clamp-only, with `--max-action-delta` unable to restore delta limiting.

---

## Fix

Restrict the rotation-representation refusal to **absolute pose modes**:

```text
_ABSOLUTE_POSE_MODES = _ABSOLUTE_MODES & _POSE_MODES = {eef_abs_pose}
```

The displacement `review()` path already clamps every dimension through the box, so no runtime change is needed there.

As a result, downstream behavior recovers automatically:

* `doctor` now reports Euler-delta embodiments as conformant.
* CLI runs retain delta limiting.

---

### `test_approvers.py`

* Kept the absolute-pose rejection (renamed to explicitly refer to "absolute").
* Added a parametrized `eef_delta_pose × {euler_xyz, quat_wxyz, axis_angle}` test asserting:

  * construction succeeds, and
  * a rotation-delta dimension clamps to the `±max_delta` box.

### `test_conformance.py`

* Added a `euler-delta-is-guardrail-conformant` test.

---

## Verification

Verified on Linux:

* ✅ `ruff`
* ✅ `ruff format`
* ✅ `mypy --strict` (62 files)
* ✅ `pytest --cov` at **100%** (`approver.py` fully covered)

---

## Result

Unblocks `inspect-robots-widowx` declaring `euler_xyz` honestly once released, per the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,15 @@ All notable changes to this project are documented here. The format is based on
   catch the `ConfigError` raised by `Task`'s epoch validation and surface it
   through the existing `_resolve_or_exit` pattern, matching how invalid
   constructor kwargs are handled for config-file components (#47).
+- **`DeltaLimitApprover` no longer rejects displacement pose modes with
+  rotation deltas** (#143). The per-dimension rotation-repr refusal now fires
+  only for **absolute** pose modes (`eef_abs_pose`), where clamping an absolute
+  euler/quat orientation has wraparound and axis-coupling problems. A
+  displacement pose mode (`eef_delta_pose`) carries small rotation deltas that
+  clamp per dimension like any other bounded displacement, so an euler-delta
+  embodiment (e.g. BridgeData V2's 7-D xyz+euler deltas) is now guardrail-ready:
+  `doctor` reports it conformant, and CLI runs keep delta limiting instead of
+  silently degrading to clamp-only.
 - **Operator scoring no longer prompts twice for self-confirming embodiments**
   (#53). On interactive ad-hoc runs, definitive `success` or `failure`
   termination verdicts are adopted as the operator judgement, announced on the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,15 +116,17 @@ All notable changes to this project are documented here. The format is based on
   catch the `ConfigError` raised by `Task`'s epoch validation and surface it
   through the existing `_resolve_or_exit` pattern, matching how invalid
   constructor kwargs are handled for config-file components (#47).
-- **`DeltaLimitApprover` no longer rejects displacement pose modes with
-  rotation deltas** (#143). The per-dimension rotation-repr refusal now fires
-  only for **absolute** pose modes (`eef_abs_pose`), where clamping an absolute
-  euler/quat orientation has wraparound and axis-coupling problems. A
-  displacement pose mode (`eef_delta_pose`) carries small rotation deltas that
-  clamp per dimension like any other bounded displacement, so an euler-delta
-  embodiment (e.g. BridgeData V2's 7-D xyz+euler deltas) is now guardrail-ready:
-  `doctor` reports it conformant, and CLI runs keep delta limiting instead of
-  silently degrading to clamp-only.
+- **`DeltaLimitApprover` no longer rejects displacement pose modes whose
+  rotation deltas are safe to clamp per dimension** (#143). The per-dimension
+  rotation-repr refusal now fires for absolute pose modes (`eef_abs_pose`,
+  where clamping an absolute orientation has wraparound and axis-coupling
+  problems) and, separately, for quaternion deltas in displacement pose modes
+  (`eef_delta_pose` + `quat_wxyz`/`quat_xyzw`, whose identity is not the zero
+  vector, so per-dimension clamping distorts the rotation instead of limiting
+  it). Euler and axis-angle deltas have no such problem and clamp fine, so an
+  euler-delta embodiment (e.g. BridgeData V2's 7-D xyz+euler deltas) is now
+  guardrail-ready: `doctor` reports it conformant, and CLI runs keep delta
+  limiting instead of silently degrading to clamp-only.
 - **Operator scoring no longer prompts twice for self-confirming embodiments**
   (#53). On interactive ad-hoc runs, definitive `success` or `failure`
   termination verdicts are adopted as the operator judgement, announced on the
--- a/docs/guide/adapters.md
+++ b/docs/guide/adapters.md
@@ -112,7 +112,7 @@ constructor hardware-free; connect in `reset()`).
 | `bounds` | Finite `low`/`high` on every dim. Without them the bounds clamp is skipped and no default delta limit can be derived. |
 | `dim_labels` | Every dim is named (`("left_j0", ..., "right_gripper")`), uniquely. The agent moves joints by these names. |
 | `state_alignment` | Absolute-target modes (`joint_pos`, `eef_abs_pose`) declare exactly one `StateSpec` field with `shape == (action_dim,)`: the proprioceptive reference the agent interpolates from. |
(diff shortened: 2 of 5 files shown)


jeqcho (changes requested):
Thanks Tushar. The test work here is the strong part: the conformance-level test pins the actual symptom from #143 (doctor reporting a BridgeData-style 7-D xyz+euler embodiment non-conformant) rather than just the constructor, and scoping the fix to the constructor gate while leaving `review()` untouched is the right cut. Defining `_ABSOLUTE_POSE_MODES` as an intersection so it self-maintains is a nice touch too.

One thing blocks it: the premise "rotation deltas clamp per dimension like any other bounded displacement" holds for `euler_xyz` and `axis_angle`, but not for quaternions. A no-op quat delta is `(1, 0, 0, 0)`, not the zero vector. Clamp `w` into a symmetric `±max_delta` box and any downstream normalization inflates the rotation instead of limiting it. Concretely: an intended delta of `(1, 0.05, 0.05, 0.05)` is about a 10 degree rotation; with `max_delta=0.05` the clamp yields `(0.05, 0.05, 0.05, 0.05)`, which normalizes to `w=0.5`, a 120 degree rotation. So `--max-action-delta`, the flag this PR restores, would amplify rotations on a quat-delta embodiment rather than limit them.

Could you keep the refusal for `quat_wxyz`/`quat_xyzw` in displacement mode (message along the lines of "quat identity is not at the origin, per-dimension clamping distorts it") and let `euler_xyz` and `axis_angle` through? The parametrized test then splits naturally: `euler_xyz`/`axis_angle` keep the happy-path clamp assertion, `quat_*` assert the refusal. Related: the current `quat_wxyz` case uses `shape=(7,)`, but a quat delta pose is 3 pos + 4 quat + 1 gripper = 8 dims, so as written it pins behavior on a layout that can't exist.

Two smaller notes, take or leave:

- `quat_xyzw` currently has no coverage either way; the split above covers it for free.
- House style: bold is reserved for definition-list lead-ins, so the mid-sentence `**absolute**` in the docstring, `adapters.md`, and the CHANGELOG body should drop to plain text.

`rot6d` in displacement mode has the same identity-not-at-origin problem (no-op is `(1,0,0,0,1,0)`), but that predates this PR, so I'll open a follow-up issue rather than load it onto this one.

The euler-delta unblocking is what #143 needed, happy to merge once the quat gate is back.

Tushar2604:
@jeqcho Thanks for the detailed review — all addressed in 1bcf88e.

1)Restored the quat refusal in displacement mode (quat_wxyz/quat_xyzw), letting euler_xyz/axis_angle through. Self-maintaining via _DISPLACEMENT_POSE_MODES = _POSE_MODES - _ABSOLUTE_POSE_MODES, mirroring the intersection pattern you liked.

2)Fixed the shape bug (quat delta pose is 8 dims, not 7) and split the test: euler_xyz/axis_angle keep the happy-path clamp assertion, quat_wxyz/quat_xyzw get their own parametrized refusal test — quat_xyzw now covered too.

3)Dropped the mid-sentence bold in the docstring, adapters.md, and the CHANGELOG, and updated the surrounding text since it no longer holds that displacement mode accepts any representation.


Verified all six mode/rep combinations end-to-end and reran the full suite on Linux: 100% coverage, mypy strict, ruff, mkdocs strict.

rot6d's identical problem in displacement mode is still out of scope here per your note — happy to pick up #150 next.

jeqcho:
The update looks right, @Tushar2604. I checked the math on the euler_xyz and axis_angle cases: both have their no-op at the origin, so a zero-centered per-dimension clip can only shrink each component, and for axis_angle that means the encoded angle (the vector norm) can only shrink too. No amplification path. Deriving `_DISPLACEMENT_POSE_MODES` from set arithmetic instead of hand-listing it also means any future pose mode lands on the unsafe-by-default side, which is the right failure direction for a guardrail. Good catch on the 8-dim quat delta shape bug in the original test data as well.

One step left before this can land: the branch now conflicts with main, but only in CHANGELOG.md (#146 and #155 merged entries under Unreleased since you branched). Could you rebase onto main and resolve that one hunk? That will also unstick CI, which GitHub silently refuses to run while a PR has conflicts, so your last two commits have never actually been through the pipeline.

Optional while you're in there: the new displacement refusal message hardcodes "euler_xyz and axis_angle are safe here" as prose. Deriving it from the sets like the absolute-mode message does with `sorted(_LIMITABLE_ROT)` would keep it from drifting if the sets ever change. Fine to skip; happy to take it as a follow-up instead.

Tushar2604:
@jeqcho Rebased onto main and resolved the CHANGELOG.md conflict — it was two overlapping additions in the same spot (my two commits both touch that Fixed entry, plus main had picked up #145 and #154 in between), nothing content-related.

Also took the optional nit: the displacement refusal message now derives its safe-rep list from get_args(RotationRepr) - _DISPLACEMENT_POSE_UNSAFE_ROT instead of hardcoding "euler_xyz and axis_angle" as prose. Folded it into the same commit that introduced the message rather than adding a third one.

Verified on Linux: ruff, format, mypy strict (63 files), 100% coverage, 743 passed. Force-pushed (0611b14) — should be conflict-free and CI should actually run now.

jeqcho (approved):
Rechecked this against the quaternion objection and it holds up now.

Ran the numbers on my own example: (1, 0.05, 0.05, 0.05) is 9.9 degrees intended, and the old per-dim clamp at max_delta=0.05 gave (0.05, 0.05, 0.05, 0.05), which normalizes to 120 degrees. 12x the wrong direction. At this head both quat orders get refused at construction instead, and the message names the off-origin identity as the reason.

The euler_xyz case from #143 does what it should: 0.5 rad on dpitch clamps to exactly 0.05 with `delta_clamped` set, in-range actions pass through untouched.

One thing to fix up whenever: the Testing section in the description is stale from before the force-push, it still says quats construct fine.

jeqcho:
Thanks Tushar. Putting the off-origin-identity reason into the error message itself means the next person to hit it gets an explanation rather than just a rejection.

Issue #47, opened 2026-07-13
config set can re-own stale [*.args]; bad kwargs from a matching section print raw tracebacks


jeqcho:
Follow-ups from the #46 review (the #44 fix is in; these are the residual gaps it documented).

## 1. `config set <kind> X` silently re-owns stale args

Ownership of an `[<kind>.args]` section is inferred at read time from the current `[defaults]` name, not recorded when the args were written. So with `[embodiment.args] rest_pose = ...` written for `yam_arms`:

```console
$ inspect-robots config set embodiment cubepick
$ inspect-robots doctor
TypeError: CubePickEmbodiment.__init__() got an unexpected keyword argument 'rest_pose'
```

The stale args now match the new owner and crash again, exactly like #44. Inherent to unscoped INI sections. Options:
- cheap: `set_default` warns when changing a component name while its `[<kind>.args]` section is non-empty
- structural: per-component scoping, `[embodiment.<name>.args]` (the alternative #44 proposed), which makes the whole class of leak impossible

## 2. `_resolve_or_exit` doesn't catch TypeError

`src/inspect_robots/cli.py` catches `KeyError`/`ConfigError` but a bad kwarg from a *matching* args section (typo'd key, wrong type) prints a raw traceback, against the CLI's own 'never print a traceback at the user' principle. Pre-existing on main; a caught `TypeError` naming the config file and section would fix the UX for case 1 as well.

## 3. Test gap: sim drop path

No test exercises `_config_args` mismatch at the sim call site (`$INSPECT_ROBOTS_SIM_EMBODIMENT` naming a different sim than the file's `sim_embodiment` while `[sim_embodiment.args]` is non-empty). The keep path and the wiring are partially pinned by existing tests; one drop-path test would finish the job.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #146, opened 2026-07-18
fix: guard --epochs 0/negative with a guided error (#145)

### LINKED ISSUE #145: [bug]  --epochs 0 (or negative) crashes with a raw traceback instead of a guided error
## Describe the bug

`--epochs N` crashes with a raw Python traceback instead of showing a guided CLI error when `N` is `0` or negative. This affects both `run` and `eval-set`.

**Root cause**

Both commands apply `--epochs` using `dataclasses.replace(task, epochs=args.epochs)` in `cli.py`.

Calling `replace()` re-runs `Task.__post_init__`, which validates the value by constructing `Epochs(count=...)`. If the count is less than `1`, it raises a `ConfigError`.

Neither call site catches that exception, so it propagates as a raw traceback.

Every other invalid CLI input (unknown `--task`, invalid `-P`/`-E` kwargs, `--sim`/`--embodiment` conflicts, etc.) is converted into a guided `SystemExit`. This is the only validation path that isn't.

This is the same bug class as **config set can re-own stale [*.args]; bad kwargs from a matching section print raw tracebacks #47**, which fixed the identical issue for config-file kwargs.

## To reproduce

```console
inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --epochs 0
```

Or without the CLI:

```python

main([
    "run",
    "--task", "cubepick-reach",
    "--policy", "scripted",
    "--embodiment", "cubepick",
    "--epochs", "0",
])
```

Both raise:

```text
...
File ".../cli.py", line 888, in _cmd_run
    task = replace(task, epochs=args.epochs)
...
File ".../task.py", line 32, in __post_init__
    raise ConfigError(f"Epochs count must be >= 1, got {self.count}")
inspect_robots.errors.ConfigError: Epochs count must be >= 1, got 0
```

`eval-set --epochs -1` follows the same code path in `_cmd_eval_set` via:

```python
[replace(t, epochs=args.epochs) for t in tasks]
```

and fails with:

```text
ConfigError: Epochs count must be >= 1, got -1
```

## Expected behavior

A non-positive `--epochs` value should exit with a guided error message that names the invalid value and how to fix it, for example:

```text
--epochs must be >= 1, got 0
```

It should not produce a traceback, matching the behavior for an unknown `--task` or invalid `-T`/`-P`/`-E` kwargs.

## Environment

- Inspect Robots version: `main @ 17c1b4f` (94 commits past `v0.8.0`)
- Python version: `3.12.4`
- OS: Windows 11 (also reproduces on Linux; this is argument validation, not platform-specific)
- NumPy version: `2.4.6`

## Additional context

Found while auditing the CLI's error-handling paths after working on **DeltaLimitApprover rejects delta pose modes with clampable rotation deltas #143**.

The fix has a clear precedent: wrap the two `replace(..., epochs=...)` call sites, catch `ConfigError`, and convert it to `SystemExit`, matching what `_resolve_or_exit` already does for the same exception type elsewhere in `cli.py`.

Since `eval-set` applies `--epochs` to each task in a list comprehension, the error should identify which task's `--epochs` value failed.

Happy to open a PR if that direction looks right.

nk-pavan-official:
Summary
--epochs 0 (or negative) crashed with a raw Python traceback instead of a clean CLI error. Fixing that.
Changes

_cmd_run: wrapped the replace(task, epochs=...) call in try/except for ConfigError, exits clean instead of crashing
_cmd_eval_set: same fix, but had to turn the list comprehension into a loop so I could catch the error per task and say which one failed
Added 2 tests in test_registry_cli.py for both call sites (0, -1, -5)
Updated after review: retyped patched as list[Task] (dropped a stale type: ignore), updated an existing test to expect SystemExit instead of the raw error, added a CHANGELOG entry

## Checklist

- [x] `mypy` passes (strict)
- [ ] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test

## Related
Closes #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,11 @@ All notable changes to this project are documented here. The format is based on
 
 ### Fixed
 
+- **`--epochs 0` or a negative value now exits with a guided error instead of a
+  raw traceback** (#145). Both `inspect-robots run` and `inspect-robots eval-set`
+  catch the `ConfigError` raised by `Task`'s epoch validation and surface it
+  through the existing `_resolve_or_exit` pattern, matching how invalid
+  constructor kwargs are handled for config-file components (#47).
 - **Operator scoring no longer prompts twice for self-confirming embodiments**
   (#53). On interactive ad-hoc runs, definitive `success` or `failure`
   termination verdicts are adopted as the operator judgement, announced on the
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -1011,8 +1011,9 @@ def _cmd_eval_set(args: argparse.Namespace) -> int:
     tasks = [_resolve_or_exit("task", name) for name in task_names]
     if args.epochs is not None:
         from inspect_robots.errors import ConfigError
+        from inspect_robots.task import Task
 
-        patched: list[Any] = []
+        patched: list[Task] = []
         for t in tasks:
             try:
                 patched.append(replace(t, epochs=args.epochs))
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -1012,13 +1012,13 @@ def _cmd_eval_set(args: argparse.Namespace) -> int:
     if args.epochs is not None:
         from inspect_robots.errors import ConfigError
 
-        patched: list[object] = []
+        patched: list[Any] = []
         for t in tasks:
             try:
                 patched.append(replace(t, epochs=args.epochs))
             except ConfigError as exc:
                 raise SystemExit(f"--epochs (task {t.name!r}): {exc}") from exc
-        tasks = patched  # type: ignore[assignment]
+        tasks = patched
 
     resolved = _resolve_components(args, defaults)
     embodiment = resolved.embodiment
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -885,7 +885,12 @@ def _cmd_run(args: argparse.Namespace) -> int:
     embodiment = resolved.embodiment
     try:
         if args.epochs is not None:
-            task = replace(task, epochs=args.epochs)
+            from inspect_robots.errors import ConfigError
+
+            try:
+                task = replace(task, epochs=args.epochs)
+            except ConfigError as exc:
+                raise SystemExit(f"--epochs: {exc}") from exc
 
         _announce_components(resolved)
         approver = _build_and_announce_guardrails(args, embodiment.info.action_space)
@@ -1005,7 +1010,15 @@ def _cmd_eval_set(args: argparse.Namespace) -> int:
     defaults = load_defaults(os.environ)
     tasks = [_resolve_or_exit("task", name) for name in task_names]
     if args.epochs is not None:
-        tasks = [replace(t, epochs=args.epochs) for t in tasks]
+        from inspect_robots.errors import ConfigError
+
+        patched: list[object] = []
+        for t in tasks:
+            try:
+                patched.append(replace(t, epochs=args.epochs))
+            except ConfigError as exc:
+                raise SystemExit(f"--epochs (task {t.name!r}): {exc}") from exc
+        tasks = patched  # type: ignore[assignment]
 
     resolved = _resolve_components(args, defaults)
     embodiment = resolved.embodiment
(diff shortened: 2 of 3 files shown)


Tushar2604:
Thanks for taking this on.

I pulled the branch and ran the actual gates rather than going off the checklist. Note: locally, mypy fails, there's no CHANGELOG diff, and an existing test fails — the checklist marks those as done, so this likely just needs a re-run after the fixes below.

mypy --strict fails
src\inspect_robots\cli.py:1021: error: Unused "type: ignore" comment [unused-ignore]
tasks was already list[Any] pre-change, so no suppression is needed. Fix: drop the ignore, retype patched as list[Task].

Breaks test_cli_run_closes_embodiment_when_validation_raises
Still asserts pytest.raises(ConfigError) for --epochs 0; this PR correctly raises SystemExit now, so it fails (1 failed, 711 passed locally). Fix: update to pytest.raises(SystemExit), then re-run the full suite since this touches a shared error path.

No CHANGELOG entry
Please add a ### Fixed entry under Unreleased (https://github.com/robocurve/inspect-robots/issues/47 is a good template).

Minor, not blocking
f"--epochs: {exc}" reads a little rough grammatically — not asking for a change.

One more ask: could you paste the before/after CLI output for run --epochs 0 and eval-set --epochs -1? Clearest confirmation it behaves as expected end to end.

@jeqcho could you take a look when you have a moment? The approach looks right to me and I ran the gates locally to flag the above, but I'll leave the actual direction/merge call to you.

nk-pavan-official:
> Thanks for taking this on.
> 
> I pulled the branch and ran the actual gates rather than going off the checklist. Note: locally, mypy fails, there's no CHANGELOG diff, and an existing test fails — the checklist marks those as done, so this likely just needs a re-run after the fixes below.
> 
> mypy --strict fails src\inspect_robots\cli.py:1021: error: Unused "type: ignore" comment [unused-ignore] tasks was already list[Any] pre-change, so no suppression is needed. Fix: drop the ignore, retype patched as list[Task].
> 
> Breaks test_cli_run_closes_embodiment_when_validation_raises Still asserts pytest.raises(ConfigError) for --epochs 0; this PR correctly raises SystemExit now, so it fails (1 failed, 711 passed locally). Fix: update to pytest.raises(SystemExit), then re-run the full suite since this touches a shared error path.
> 
> No CHANGELOG entry Please add a ### Fixed entry under Unreleased (#47 is a good template).
> 
> Minor, not blocking f"--epochs: {exc}" reads a little rough grammatically — not asking for a change.
> 
> One more ask: could you paste the before/after CLI output for run --epochs 0 and eval-set --epochs -1? Clearest confirmation it behaves as expected end to end.
> 
> @jeqcho could you take a look when you have a moment? The approach looks right to me and I ran the gates locally to flag the above, but I'll leave the actual direction/merge call to you.

nk-pavan-official:
Fixed all three. Retyped patched as list[Task] and dropped the stale type: ignore, updated the test to expect SystemExit with a message check, and added a CHANGELOG entry under Fixed matching #47's format. Full suite passes now - 728 passed, 2 skipped (unrelated, missing rerun-sdk), 100% coverage, mypy clean.

```
$ inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --epochs 0
--epochs: Epochs count must be >= 1, got 0

$ inspect-robots eval-set --task cubepick-reach --policy scripted --embodiment cubepick --epochs -1
--epochs (task 'cubepick-reach'): Epochs count must be >= 1, got -1
```

jeqcho (approved):
Thanks Pavan, this is ready. A few things stood out beyond the fix itself.

The test coverage is thorough where it counts: parametrized over 0 and negatives at both call sites, asserting the flag name, the offending value, and the task name all show up in the message. Strengthening `test_cli_run_closes_embodiment_when_validation_raises` with a message assertion was a good catch too, since that test guards the embodiment-close invariant and was previously asserting the buggy behavior.

Also appreciated how you handled review: all three findings fixed within hours, with before/after CLI output pasted as proof, and a CHANGELOG entry matching the #47 format.

Merging. One cosmetic thing I'll fold into a follow-up rather than block on: `--epochs: Epochs count must be >= 1, got 0` doubles up a bit; something like `--epochs must be >= 1, got 0` reads cleaner. There is also an open idea to extract the two inline try/excepts into a shared helper next to `_resolve_or_exit`. If either interests you, they are yours.

jeqcho:
Merged, thanks Pavan. The follow-ups from my review are now written up in #152 if you want a second one; it builds directly on what you landed here. Either way, glad to have you contributing.

nk-pavan-official:
Thanks Jay, appreciate it. Will check out #152, that refactor idea sounds fun

PR #148, opened 2026-07-19
fix(cli): guided error for --epochs 0/negative instead of raw traceback

### LINKED ISSUE #145: [bug]  --epochs 0 (or negative) crashes with a raw traceback instead of a guided error

ptlnextdoor:
## Summary

`--epochs 0` (or negative) crashes both `run` and `eval-set` with a raw traceback instead of the guided `SystemExit` every other invalid CLI input gets.

`replace(task, epochs=...)` reruns `Task.__post_init__`, which raises `ConfigError` when the count is below 1. Neither `_cmd_run` nor `_cmd_eval_set` caught it — same bug class already fixed for config-file kwargs in #47.

## Changes

- Adds `_apply_epochs_or_exit`, mirroring the existing `_resolve_or_exit` pattern: catches `ConfigError` and re-raises as `SystemExit(str(exc))`.
- Both call sites (`_cmd_run`, `_cmd_eval_set`) now go through it instead of a bare `dataclasses.replace`.
- Drops the now-unused local `replace` imports in both functions.
- Updates `test_cli_run_closes_embodiment_when_validation_raises` (was asserting the raw `ConfigError` leaked, which was the bug) and adds `test_cli_eval_set_negative_epochs_is_a_guided_error` for the `eval-set` path, which had no coverage before.

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run ruff format --check .`
- [x] `uv run mypy` (strict)
- [x] `uv run pytest --cov` — 724 passed, 100% line+branch coverage
- [x] `uv run pre-commit run --all-files`
- [x] Manually confirmed `inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --epochs 0` now exits with the guided message instead of a traceback

Fixes #145

### DIFF
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -71,6 +71,7 @@ if TYPE_CHECKING:
     from inspect_robots.rollout import TrialRecord
     from inspect_robots.scene import Scene
     from inspect_robots.spaces import Box
+    from inspect_robots.task import Task
 
 
 def _styled(text: str, code: str) -> str:
@@ -478,6 +479,23 @@ def _resolve_or_exit(
         ) from exc
 
 
+def _apply_epochs_or_exit(task: Task, epochs: int) -> Task:
+    """Apply ``--epochs`` with a guided error instead of a raw traceback (#145).
+
+    ``replace()`` reruns ``Task.__post_init__``, which rejects a count below 1
+    via ``ConfigError`` — the same validation-error class ``_resolve_or_exit``
+    already converts to ``SystemExit`` for registry resolution.
+    """
+    from dataclasses import replace
+
+    from inspect_robots.errors import ConfigError
+
+    try:
+        return replace(task, epochs=epochs)
+    except ConfigError as exc:
+        raise SystemExit(str(exc)) from exc
+
+
 def _build_guardrails(
     space: Box, max_action_delta: float | None
 ) -> tuple[Approver, list[str], list[str]]:
@@ -835,8 +853,6 @@ def _build_and_announce_guardrails(args: argparse.Namespace, action_space: Box) 
 
 
 def _cmd_run(args: argparse.Namespace) -> int:
-    from dataclasses import replace
-
     from inspect_robots import eval
     from inspect_robots.logging import JsonLogSink
     from inspect_robots.scene import Scene
@@ -885,7 +901,7 @@ def _cmd_run(args: argparse.Namespace) -> int:
     embodiment = resolved.embodiment
     try:
         if args.epochs is not None:
-            task = replace(task, epochs=args.epochs)
+            task = _apply_epochs_or_exit(task, args.epochs)
 
         _announce_components(resolved)
         approver = _build_and_announce_guardrails(args, embodiment.info.action_space)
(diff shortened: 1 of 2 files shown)


jeqcho:
Thanks for this ptlnextdoor, and sorry for the awkward outcome: #146 was opened a day earlier by the contributor who had claimed #145 in the issue thread, so it got priority and just merged. Closing this one as a duplicate.

That said, your PR got real things right that the merged one didn't. `_apply_epochs_or_exit` mirroring `_resolve_or_exit` is the shape this code should have, down to the import hygiene (moving `replace` into the helper, `Task` under `TYPE_CHECKING`). The `pytest.raises(match=...)` exact-message pinning is stronger than the substring asserts that landed. And you correctly spotted that the old close-embodiment test was asserting the bug itself.

I opened #152 to bring exactly those pieces into main, and your branch is most of the way there already. If you want to send it, it's yours; comment on #152 to claim it. Hope to see more from you either way.

PR #153, opened 2026-07-21
Fix/delta limit rot6d displacement

### LINKED ISSUE #150: [bug] DeltaLimitApprover can amplify rot6d rotation deltas in displacement pose modes
## Describe the bug

`DeltaLimitApprover` treats `_LIMITABLE_ROT = {none, rot6d}` as safe to clamp per dimension in every pose mode. For absolute orientations that is the point of the set. But in a displacement pose mode (`eef_delta_pose`), a `rot6d` value is a rotation whose no-op is `(1, 0, 0, 0, 1, 0)`, not the zero vector.

Per-dimension clamping toward a symmetric `±max_delta` box drags the identity columns toward zero. After the Gram-Schmidt orthonormalization every consumer applies, the result can be an arbitrary large rotation, so the delta limiter can amplify a rotation instead of limiting it. Same failure class as clamping quaternion deltas, which #144 review flagged: any representation whose identity is not at the origin cannot be box-clamped safely.

## Expected behavior

The limitable set should be mode-dependent rather than a single frozenset:

- absolute pose modes: `{none, rot6d}` (current behavior, correct)
- displacement pose modes: `{none, euler_xyz, axis_angle}`, refusing `rot6d` and `quat_*` with a message explaining the identity-not-at-origin problem

## Notes

Pre-existing behavior, not a regression: `eef_delta_pose` with `rot6d` already reached the displacement clamp path before #144. Fixing it is a behavior change for any embodiment currently declaring rot6d deltas, so it needs a CHANGELOG entry and a conformance test alongside the approver unit tests.

Tushar2604:
# Fix unsafe `rot6d` displacement pose handling in `DeltaLimitApprover`

## Summary

`DeltaLimitApprover` incorrectly allowed `rot6d` in displacement pose mode (`eef_delta_pose`).

Unlike translation deltas, a `rot6d` "no-op" is **not** the zero vector—it is `(1, 0, 0, 0, 1, 0)`. Clamping each dimension independently toward a symmetric `±max_delta` range distorts this representation. Since every downstream consumer re-normalizes `rot6d` using Gram-Schmidt, the distorted value can become an arbitrarily large rotation.

As a result, the delta limiter can **amplify** rotations instead of limiting them.

This is the same underlying failure mode as clamping absolute quaternions, and the same class of issue already addressed for `quat_wxyz`/`quat_xyzw` displacement poses in **#144**.

This is **not a regression**—the behavior existed before #143 and #144.

---

### Validation

- Added `"rot6d"` to `_DISPLACEMENT_POSE_UNSAFE_ROT`.
  - The set is now:
    ```python
    {quat_wxyz, quat_xyzw, rot6d}
    ```
- No additional logic was required.
  - The existing constructor validation and error message already handle all entries in this set.
- `rot6d` **remains supported** for absolute orientation modes.
  - Only displacement (`eef_delta_pose`) is rejected.

### Tests

- Generalized the displacement validation test to accept `(rotation_repr, rot_dim)` pairs.
  - This allows testing both quaternion (4D) and `rot6d` (6D) action spaces.
- Added a conformance test verifying that `eef_delta_pose + rot6d` now raises the expected guardrails error.
  - This mirrors the existing coverage for:
    - valid Euler displacement
    - invalid absolute quaternion

### Documentation

- Updated `adapters.md` to describe the guardrail based on the actual failure condition instead of listing individual rotation representations.
- Added a **breaking** entry under **Unreleased → Fixed** in `CHANGELOG.md`.

---

## Why this change?

`rot6d` cannot be safely limited using independent per-dimension clamping in displacement mode because:

1. Its identity rotation is **not** the origin.
2. Clamping distorts the representation.
3. Gram-Schmidt normalization projects the distorted vector back onto SO(3).
4. The resulting rotation can be much larger than the intended delta.

Rejecting this configuration is therefore the correct and safe behavior.

---

## Checklist

- [x] Added unit and conformance tests
- [x] `ruff check .`
- [x] `ruff format --check .`
- [x] `mypy` (strict)
- [x] Full test suite passes
  - 730 tests
  - 100% coverage
- [x] Updated `CHANGELOG.md`
- [x] No public API changes
- [x] No new dependencies (NumPy-only)

> **Note:** `pre-commit run --all-files` could not be executed locally because it shells out to `uv`, which was unavailable. The equivalent checks (`ruff`, `mypy`, `pytest --cov`) were run directly, along with the full Linux test suite.

---

## Related

**Closes #150**

This branch is based on **#144**, since it builds directly on the `_DISPLACEMENT_POSE_MODES` and `_DISPLACEMENT_POSE_UNSAFE_ROT` infrastructure introduced there. It will be rebased onto `main` once #144 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -137,6 +137,16 @@ All notable changes to this project are documented here. The format is based on
 
 ### Fixed
 
+- **`DeltaLimitApprover` refuses `rot6d` rotation deltas in displacement pose
+  modes** (#150, breaking for any embodiment currently declaring them). A
+  `rot6d` delta's identity is `(1, 0, 0, 0, 1, 0)`, not the zero vector, so
+  per-dimension clamping toward a symmetric `±max_delta` box drags it away
+  from identity; the Gram-Schmidt re-normalization every consumer applies can
+  then amplify the rotation instead of limiting it — the same failure class
+  as clamping an absolute quaternion, and pre-existing behavior rather than a
+  regression (`eef_delta_pose` + `rot6d` already reached the displacement
+  clamp path before #143/#144). `euler_xyz` and `axis_angle` deltas have no
+  such problem and remain guardrail-conformant.
 - **An explicit invalid `--max-action-delta` now fails fast instead of silently
   running with weaker guardrails** (#154). Non-finite or non-positive values
   were previously caught by `_build_guardrails`'s degrade-per-component path
--- a/docs/guide/adapters.md
+++ b/docs/guide/adapters.md
@@ -112,7 +112,7 @@ constructor hardware-free; connect in `reset()`).
 | `bounds` | Finite `low`/`high` on every dim. Without them the bounds clamp is skipped and no default delta limit can be derived. |
 | `dim_labels` | Every dim is named (`("left_j0", ..., "right_gripper")`), uniquely. The agent moves joints by these names. |
 | `state_alignment` | Absolute-target modes (`joint_pos`, `eef_abs_pose`) declare exactly one `StateSpec` field with `shape == (action_dim,)`: the proprioceptive reference the agent interpolates from. |
-| `guardrails` | `DeltaLimitApprover(action_space)` constructs. This catches absolute pose modes (`eef_abs_pose`) whose rotation representation cannot be clamped per dimension (`quat_*`, `axis_angle`, `euler_xyz`; use `none` or `rot6d`), and displacement pose modes (`eef_delta_pose`) carrying a quaternion delta (`quat_wxyz`, `quat_xyzw`; use `none`, `rot6d`, `axis_angle`, or `euler_xyz`). |
+| `guardrails` | `DeltaLimitApprover(action_space)` constructs. This catches absolute pose modes (`eef_abs_pose`) whose rotation representation cannot be clamped per dimension (`quat_*`, `axis_angle`, `euler_xyz`; use `none` or `rot6d`), and displacement pose modes (`eef_delta_pose`) whose rotation representation's identity is not the zero vector (`quat_*`, `rot6d`; use `none`, `axis_angle`, or `euler_xyz`). |
 
 ### Warnings
 
--- a/docs/guide/policies-and-embodiments.md
+++ b/docs/guide/policies-and-embodiments.md
@@ -25,7 +25,7 @@ class MyVLA:
             action_space=Box(
                 shape=(7,),
                 semantics=ActionSemantics(
-                    control_mode="eef_delta_pose", rotation_repr="rot6d", gripper="continuous"
+                    control_mode="eef_delta_pose", rotation_repr="euler_xyz", gripper="continuous"
                 ),
             ),
             observation_space=ObservationSpace(
--- a/src/inspect_robots/approver.py
+++ b/src/inspect_robots/approver.py
@@ -84,7 +84,8 @@ _POSE_MODES = frozenset({"eef_abs_pose", "eef_delta_pose"})
 # dimension has wraparound and axis-coupling problems, so those reps are
 # refused. Displacement pose modes (eef_delta_pose) carry small rotation
 # *deltas*, which mostly clamp per dimension like any other bounded
-# displacement — except quaternions (see _DISPLACEMENT_POSE_UNSAFE_ROT below).
+# displacement — except reps whose no-op is not the origin (see
+# _DISPLACEMENT_POSE_UNSAFE_ROT below).
 _ABSOLUTE_POSE_MODES = _ABSOLUTE_MODES & _POSE_MODES
 _DISPLACEMENT_POSE_MODES = _POSE_MODES - _ABSOLUTE_POSE_MODES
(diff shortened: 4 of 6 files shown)


jeqcho:
Reviewed alongside #144 and the rot6d reasoning holds up: the no-op at (1, 0, 0, 0, 1, 0) sits off-origin, so box-clamping toward zero distorts it and Gram-Schmidt downstream can turn that into an arbitrary rotation. Adding rot6d to `_DISPLACEMENT_POSE_UNSAFE_ROT` while keeping it clampable for absolute modes is exactly what #150 asked for, and the conformance tests in `tests/test_conformance.py` closing the loop through `check_embodiment` rather than stopping at constructor behavior is appreciated.

Since this branch carries #144's commits, the path to landing is: #144 rebases and merges first, then rebase this one onto main so the diff reduces to the rot6d commit. Right now both branches share the same one-hunk CHANGELOG.md conflict with main, and CI won't run on either until the conflict clears (GitHub skips pull_request workflows when it can't build the merge commit).

Tushar2604:
Understood on the sequencing — #144 is rebased and pushed (conflict-free now, force-pushed as 0611b14), so it's ready on your end. I won't rebase this branch until #144 actually lands on main, since rebasing onto my own unmerged branch right now wouldn't be the same as rebasing onto main, and could drift if anything changes during merge. Will rebase this one as soon as #144 is in.

jeqcho:
#144 is on main now, so this is clear to rebase whenever you get a minute.

Tushar2604:
Rebased onto main using `git rebase --onto` to skip the two commits already merged via #144 (they now conflict directly since #144 was itself rebased) — diff reduces to 1 commit, 5 files, just the rot6d work, as predicted.

The `_DISPLACEMENT_SAFE_ROT` derivation (#144's optional nit) wasn't in this branch yet; kept it, and it picked up rot6d automatically — the message now reads `only ['axis_angle', 'euler_xyz', 'none'] are safe here` with no manual edit needed.

Verified on Linux: 100% coverage, mypy strict, ruff, 756 passed. Force-pushed (`3601e18`), conflict-free.

jeqcho (approved):
Approved on the merits. One small sweep and I'll merge.

I measured the amplification rather than taking it on faith. A rot6d delta of (1, 0.02, 0, 0, 1, 0.02) clamped per-dimension to ±0.05 and run through Gram-Schmidt comes out at 31.8 degrees against 1.62 intended, so about 20x, in the direction a limiter is supposed to prevent. Worth noting for the record that a pure identity no-op does survive clamping, since both columns scale uniformly and decode to 0 degrees. It's the off-axis components in any real policy output that make it bite, which is if anything a better argument than the one in the description.

Exactly one cell of the mode x repr matrix moves: `eef_delta_pose` + `rot6d` goes accept to reject, and the other twelve are identical to main including `eef_abs_pose` + `rot6d`. Nothing in-tree declares `eef_delta_pose` at all, and out of tree `_build_guardrails` degrades with a warning rather than blocking, so the failure mode for anyone affected is losing the delta limiter rather than losing the run. Given the 20x, that's the safer side.

Three bits of stale text to fold in:

- `approver.py:132-134`, the class docstring still says construction refuses a displacement pose mode carrying a quaternion delta. rot6d is refused now too, and that docstring renders into the API docs.
- `tests/test_approvers.py:64-65`, the comment points readers at `test_refuses_displacement_pose_mode_with_quat_rotation`, which this PR renamed to `..._off_origin_rotation`.
- `docs/guide/policies-and-embodiments.md:28`, the `MyVLA` example declares `control_mode="eef_delta_pose", rotation_repr="rot6d"`. That's the combo this turns into a conformance error, so as it stands we'd ship a guide showing a config that now fails. adapters.md got swept, this one got missed. While you're in there, that example's `shape=(7,)` can't hold 3 pos + 6 rot6d + 1 gripper either, if you want to fix it.

Ping me when it's pushed and I'll take it in.

Tushar2604:
Hi @jeqcho , Thanks for the review , I have done the changes.

All three folded in and pushed (`efe8c0e`):

- `approver.py` docstring now says "quaternion or rot6d delta".
- `test_delta_pose_rotation_deltas_clamp_per_dim`'s comment now points at `test_refuses_displacement_pose_mode_with_off_origin_rotation`.
- `policies-and-embodiments.md`'s `MyVLA` example switched to `rotation_repr="euler_xyz"`, which also fixes the shape — `shape=(7,)` is correct for 3 pos + 3 euler + 1 gripper, so no dim change needed.

Also folded the measured 20x amplification number into the commit message since it's a stronger argument than what I originally wrote.

Verified on Linux: 100% coverage, mypy strict, ruff, 756 passed. Ready when you are.

jeqcho:
Merged. Picking `euler_xyz` for the MyVLA example was the right instinct: it gives 3 pos + 3 euler + 1 gripper, which is the `shape=(7,)` the example already declared. The old rot6d version needed 10, so that example had been dimensionally wrong the whole time and nobody noticed. Nice catch on the way past.

Also appreciated you holding the rebase until #144 actually landed rather than rebasing onto your own unmerged branch. That was the right read.

PR #155, opened 2026-07-21
fix(cli): guard --max-action-delta against invalid explicit values

### LINKED ISSUE #154: [bug] An explicit invalid --max-action-delta silently weakens guardrails instead of erroring
**Describe the bug**

An invalid explicit `--max-action-delta` value silently weakens guardrails instead of failing. Instead of rejecting invalid user input, the CLI prints a warning to `stderr`, skips the delta limiter, continues with clamp-only guardrails, and exits successfully (`0`).

**Root cause:**
- `_build_guardrails` (`cli.py`) wraps `DeltaLimitApprover(space, max_delta=max_action_delta)` in a blanket `except ValueError`, and degrades to a warning for any failure.
- That degrade-per-component design is correct when a *derived* default can't apply to a space (for example, an embodiment declares no bounds), because there was never a working default to lose.
- However, when the user explicitly passes `--max-action-delta`, an invalid value (negative, zero, or non-finite) is a CLI input error, not an embodiment limitation. `DeltaLimitApprover` correctly rejects the value (`max_delta must be finite and > 0`), but that exception is caught by the same blanket handler and downgraded to a warning.
- As a result, the operator asked for a stricter per-step limit, mistyped it, and the run proceeds on real hardware with **less protection than explicitly requested**, unless they happen to notice the warning on `stderr`.

**To reproduce**

```console
inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --max-action-delta -1
```

Or without the CLI:

```python

main([
    "run",
    "--task", "cubepick-reach",
    "--policy", "scripted",
    "--embodiment", "cubepick",
    "--max-action-delta", "-1",
])
```

Output:

```text
guardrails warning: delta limit skipped: DeltaLimitApprover: max_delta must be finite and > 0
policy: scripted (--policy)
embodiment: cubepick (--embodiment)
guardrails: clamp
run status: completed
...
```

Exit code: `0`.

The run completes with **clamp-only** guardrails instead of the intended **clamp + delta-limit** chain configured by the flag.

**Expected behavior**

An explicitly invalid `--max-action-delta` (non-finite or `<= 0`) should fail immediately before any component resolution or hardware interaction, with a clear error such as:

```text
--max-action-delta must be finite and > 0, got -1.0
```

This should **not** change the existing degrade-per-component behavior for derived defaults. An embodiment with no declared bounds and no explicit `--max-action-delta` should continue warning and proceeding as it does today. Only an explicitly supplied invalid value should hard-fail.

**Environment**
- Inspect Robots version: `main` @ `7992f14` (`v0.21.1`, 4 commits past tag)
- Python version: 3.12.4
- OS: Windows 11 (also reproducible on Linux; this is argument validation, not platform-specific)
- NumPy version: 2.4.6

**Additional context**

- `argparse` uses `type=float` for `--max-action-delta`, so the CLI value is always a `float | None`. From the CLI, the only reachable validation failures are non-finite values or values `<= 0`. The broadcast-shape validation in `DeltaLimitApprover._validate_max_delta()` is only reachable through the Python API by passing an array-like value directly.
- A narrow fix would be to validate `args.max_action_delta` in `_check_shared_run_conflicts()` (`cli.py`). That function already runs before component resolution for both `run` and `eval-set`, and already contains the sibling `--disable-guardrails` / `--max-action-delta` conflict check.
- No changes to `_build_guardrails` or `DeltaLimitApprover` itself should be necessary.

Tushar2604:
## Summary

An explicit invalid `--max-action-delta` silently weakened guardrails instead of erroring.

The bug:
- `_build_guardrails` wraps `DeltaLimitApprover(space, max_delta=max_action_delta)` in a blanket `except ValueError` and degrades to a stderr warning for any failure.
- That degrade-per-component design is correct for a *derived* default an embodiment's space can't support — there was never a working default to lose.
- It is not correct for an explicitly passed, malformed value (negative, zero, non-finite). That is a CLI input error, not an embodiment limitation, but it hit the same blanket handler and got downgraded to a warning.
- Result: the operator asks for a tighter per-step limit, mistypes it, and the run proceeds on real hardware with clamp-only guardrails and exit code `0` — silently, unless they are watching stderr.

## Changes

- Validate `args.max_action_delta` in `_check_shared_run_conflicts`, which already runs before component resolution in both `run` and `eval-set`, and already houses the sibling `--disable-guardrails` conflict check. A non-finite or non-positive value now exits with a guided `SystemExit` before anything resolves or energizes.
- No change needed in `_build_guardrails` or `DeltaLimitApprover` — `argparse`'s `type=float` means the CLI value is always a plain float, so the other failure mode in `_validate_max_delta` (array broadcast) can only be reached via the Python API, not this flag.
- Derived-limit degradation is untouched: an embodiment with no declared bounds and no explicit `--max-action-delta` still warns and continues, exactly as before. Verified manually and confirmed the existing `test_cli_degraded_guardrails_warn_but_run` still passes unchanged.
- Added parametrized tests for both `run` and `eval-set` (`-1`, `0`, `inf`, `nan`).
- Added a `CHANGELOG.md` entry under `Unreleased` → `Fixed`.

## Checklist

- [x] pre-commit hooks pass — `pre-commit run --all-files` shells out to `uv`, not on PATH here; ran `ruff check .`, `ruff format --check .`, `mypy`, and `pytest --cov` directly, plus the full suite on a Linux checkout.
- [x] Tests added/updated — 8 new parametrized cases (4 values × run/eval-set).
- [x] Coverage stays at 100% — verified on Linux: 100.00% total, 738 passed.
- [x] `mypy` passes (strict) — full run, 63 source files.
- [x] `CHANGELOG.md` updated under "Unreleased" — `Fixed` entry.
- [x] Public API unaffected — `_check_shared_run_conflicts` is a private CLI function.
- [x] Core stays NumPy-only — no new deps.

Note: I did not run the `docs-build` job locally (it needs a full Node/npm toolchain for the Docusaurus site). Confirmed by inspection this diff is out of scope for it — no changes to `docs/`, `website/`, or any public docstring, and `CHANGELOG.md` is not referenced by the site build.

## Related

Closes #154.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,16 @@ All notable changes to this project are documented here. The format is based on
 
 ### Fixed
 
+- **An explicit invalid `--max-action-delta` now fails fast instead of silently
+  running with weaker guardrails** (#154). Non-finite or non-positive values
+  were previously caught by `_build_guardrails`'s degrade-per-component path
+  (meant for derived limits an embodiment's space can't support) and
+  downgraded to a stderr warning, so the run proceeded with clamp-only
+  guardrails despite the operator explicitly asking for a tighter limit. Both
+  `run` and `eval-set` now reject a malformed explicit value in the shared
+  conflict check, before anything resolves or energizes. Derived-limit
+  degradation (no explicit flag, an embodiment declaring no bounds) is
+  unaffected — it still warns and continues.
 - **`--epochs 0` or a negative value now exits with a guided error instead of a
   raw traceback** (#145). Both `inspect-robots run` and `inspect-robots eval-set`
   catch the `ConfigError` raised by `Task`'s epoch validation and surface it
--- a/src/inspect_robots/cli.py
+++ b/src/inspect_robots/cli.py
@@ -745,7 +745,7 @@ class _ResolvedComponents(NamedTuple):
 
 
 def _check_shared_run_conflicts(args: argparse.Namespace) -> None:
-    """Reject flag combinations invalid for both ``run`` and ``eval-set``."""
+    """Reject flag combinations and malformed values invalid for both ``run`` and ``eval-set``."""
     if args.sim and args.embodiment:
         raise SystemExit(
             "--sim selects your configured sim_embodiment; "
@@ -755,6 +755,15 @@ def _check_shared_run_conflicts(args: argparse.Namespace) -> None:
         raise SystemExit(
             "--max-action-delta tunes the guardrails that --disable-guardrails turns off — drop one"
         )
+    if args.max_action_delta is not None and not (
+        math.isfinite(args.max_action_delta) and args.max_action_delta > 0
+    ):
+        # A malformed *explicit* value is a CLI input error, not an embodiment
+        # limitation: fail fast here rather than let _build_guardrails's
+        # degrade-per-component path (meant for derived limits an embodiment's
+        # space can't support) downgrade it to a warning and run with weaker
+        # guardrails than the operator explicitly asked for.
+        raise SystemExit(f"--max-action-delta must be finite and > 0, got {args.max_action_delta}")
 
 
 def _resolve_components(args: argparse.Namespace, defaults: Defaults) -> _ResolvedComponents:
--- a/tests/test_registry_cli.py
+++ b/tests/test_registry_cli.py
@@ -2835,6 +2835,36 @@ def test_cli_max_action_delta_conflicts_with_disable(tmp_path: Path) -> None:
         )
 
 
+@pytest.mark.parametrize("value", ["-1", "0", "inf", "nan"])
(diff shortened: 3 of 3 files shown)


jeqcho:
Merged, thank you @Tushar2604. Two things stood out. Putting the guard in `_check_shared_run_conflicts` instead of `_build_guardrails` was the right architectural read: the catch-and-warn path exists for derived defaults an embodiment can't support, and your fix closes the explicit-input hole without touching that behavior (leaving `test_cli_degraded_guardrails_warn_but_run` untouched made that easy to confirm). And the parametrized tests covering -1, 0, inf, and nan on both `run` and `eval-set` pin the contract exactly where the bug lived.

Good hunting on this one. More where it came from is always welcome.

PR #161, opened 2026-07-22
feat: add seconds-based task horizons

### LINKED ISSUE #160: [design] Seconds-based task horizons resolved against the embodiment's control rate at eval() time
## Problem

`Task.max_steps` is an integer step budget fixed at task construction time, but a step only has a duration once an embodiment is paired: the same 600-step horizon is 60s on a 10 Hz mock world and 40s on a 15 Hz arm. Benchmarks that specify physical time budgets (kitchenbench's protocol allots per-task seconds; see robocurve/kitchenbench#28 and the discussion on robocurve/kitchenbench#29) currently have no way to express them. Task factories behind the string registry are zero-arg, so a benchmark cannot thread the paired embodiment's rate into `make_task` itself, and `Task` is deliberately embodiment-agnostic (R1: the rollout enforces no wall-clock rate; `TaskEnvelope` carries no control rate on purpose).

## Sketch

Let a task declare its horizon in physical time, and let `eval()` resolve it against the pair:

- `Task` grows an optional `max_seconds: float | None` (mutually exclusive with `max_steps`).
- `eval()` resolves `max_steps = ceil(max_seconds * embodiment.info.control_hz)` after compatibility checking, before the rollout, and stamps both the resolved value and the declared seconds into `EvalLog` metadata so runs at different rates remain interpretable.
- `TaskEnvelope.max_steps` keeps its current meaning (adapters see the resolved integer), so `bind_task` consumers are unaffected.

This does not reintroduce pacing: the rollout still never sleeps; `control_hz` is only used as a conversion factor for the budget, which is exactly the quantity `info.control_hz` already documents.

## Open questions

- What should happen when `info.control_hz` is None or zero (event-driven embodiments)? Refuse at compatibility-check time, or require `max_steps` for such pairs?
- Should scorers see the declared seconds (for time-normalized metrics) or stay steps-only?
- Does `eval-set` need per-pair resolved horizons surfaced in the comparison table?

Found while reviewing robocurve/kitchenbench#29, which attempted this benchmark-side and could not reach the paired embodiment's rate through the zero-arg registry path.

Vedangalle:
## Summary

Add seconds-based task horizons as an alternative to fixed step counts. This lets benchmark authors express a physical duration without coupling the task definition to one embodiment's control rate, while preserving the integer `TaskEnvelope.max_steps` contract used by adapters.

## Changes

- Require each `Task` to declare exactly one of `max_steps` or keyword-only `max_seconds`.
- Validate seconds horizons and embodiment `control_hz` values as positive and finite.
- Resolve `ceil(max_seconds * control_hz)` after compatibility checking and before `bind_task()` or rollout, with a one-step minimum for positive floating-point underflow.
- Record the declared `max_seconds` and resolved `max_steps` in EvalLog schema v1 while preserving old-log compatibility.
- Show declared and resolved horizons in CLI notices, `inspect`, `eval-set`, and HTML reports.
- Document the design and migration behavior; update the changelog and edge-case coverage.

Rate-unknown and event-driven embodiments continue to use `max_steps`. Scorer interfaces, ad-hoc `--max-steps`, rollout pacing, and dependencies are unchanged.

## Checklist

- [x] `pre-commit run --all-files` and the pre-push hook stage pass.
- [x] Tests added/updated, including mock-world orchestration and edge cases.
- [x] Coverage remains at **100%** line and branch (`769 passed, 2 skipped`).
- [x] `ruff check .` and `ruff format --check .` pass.
- [x] Strict `mypy` passes.
- [x] `CHANGELOG.md` updated under "Unreleased".
- [x] Public API behavior is documented and API snapshot tests pass; no new exported symbol is required.
- [x] Core remains NumPy-only with no new dependencies.

Additional documentation checks:

- API reference generation passes.
- Docusaurus production build passes.

## Related

Closes #160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ All notable changes to this project are documented here. The format is based on
 
 ### Changed
 
+- **Task horizon binding now follows compatibility checking.** An
+  embodiment's optional `bind_task()` hook receives the resolved step envelope
+  only after the policy/embodiment/task triple is known to be compatible. This
+  prevents adapters from acting on a seconds-derived budget built from an
+  invalid control rate (#160).
+
 - **Docs site migrated from MkDocs Material to Docusaurus.** The site at
   inspectrobots.org now builds from `website/` (Docusaurus 3) while the
   Markdown source stays in `docs/`; every existing URL, `llms.txt`, and
@@ -38,6 +44,13 @@ All notable changes to this project are documented here. The format is based on
   one `max_llm_calls` unit.
 
 ### Added
+
+- **Seconds-based benchmark horizons:** `Task(max_seconds=...)` gives every
+  compatible embodiment the same physical-time budget. `eval()` resolves it
+  with `ceil(max_seconds * embodiment.info.control_hz)`, rejects missing or
+  invalid control rates before `bind_task()` or rollout, and records both the
+  declared seconds and resolved steps in eval logs, CLI summaries, inspection,
+  and HTML reports (#160).
 
 - **Policy lifecycle hook: `on_trial_end`** — policies can now hook into
   the end of a trial to persist state or artifacts. The orchestrator calls
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -210,6 +210,10 @@ inspect-robots eval-set 'kitchenbench/*' --policy xpolicylab -P url=ws://host:19
 inspect-robots eval-set cubepick-reach my-other-task --policy scripted --embodiment cubepick
 ```
 
+For a task declared with `max_seconds`, its summary row includes both the
+physical-time budget and the integer step limit resolved from the selected
+embodiment, for example `[120s -> 1200 steps at 10 Hz]`.
+
 Multiple patterns may match the same task; it still runs once. A pattern that
 matches nothing is an error listing every registered task. `--policy` and
 `--embodiment` (and `-P`/`-E`, `--sim`, `--epochs`, `--fail-on-error`,
@@ -253,6 +257,7 @@ policy:      scripted
 embodiment:  cubepick
 run status:  completed
 outcome:     5 succeeded
+horizon:     120s -> 1200 steps at 10 Hz
 scenes:      5   trials: 5
 metrics:
   success_at_end: 1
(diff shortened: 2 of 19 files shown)


Adityakk9031:
please close this maintainer assigned this issue to me

Adityakk9031:
check this https://github.com/robocurve/kitchenbench/pull/29

Vedangalle:
Thanks for flagging this, and sorry - I missed Jay’s note on KitchenBench #29 that #160 was yours first. I saw the framework issue unassigned and implemented it from there; I don’t want to duplicate or cut across your work.

Since #161 now contains the core framework implementation and validation, could we coordinate by keeping this PR for the `inspect-robots` hook while you continue the KitchenBench task and version migration in #29 against it? @jeqcho, I’m happy to follow your preference if you want the ownership structured differently.

Adityakk9031:
please close this

Adityakk9031:
@Vedangalle  please close this pr

jeqcho:
Hi @Vedangalle and @Adityakk9031, thanks both for taking #160 on.

I'm going with #161. To be clear about why, since the sequencing is messy: I offered this to @Adityakk9031 on [kitchenbench#29](https://github.com/robocurve/kitchenbench/pull/29) on 07-21, and I should have assigned the issue then. That one's on me. But the reply came 21 hours later, and by that point #161 by @Vedangalle had been up and ready for review (07-22 07:55:34 UTC (https://github.com/robocurve/inspect-robots/pull/161#event-28313355363), while @Adityakk9031 claimed it 11 hours after #161 at 19:30:21 (https://github.com/robocurve/kitchenbench/pull/29#issuecomment-5050553061)). I'd rather not discard finished work over a claim that arrived after it.

To @Vedangalle: the design doc helped, and two calls in particular: making max_seconds keyword-only so the positional Task(...) signature doesn't shift under existing benchmarks, and moving assert_compatible ahead of bind_task so a bad control_hz is caught before an adapter sees the envelope.

To @Adityakk9031: your PR #162 does one thing #161 doesn't: it rejects bool for max_steps, so Task(max_steps=True) can't slip through. This is good, and I would love to see a PR from you on this. There are plenty other issues open across github.com/robocurve if you want the next one.

jeqcho:
Merged. The underflow floor is the detail I liked most here: `max(1, ceil(...))` means a tiny `max_seconds` on a slow embodiment still gets one step instead of a trial that ends before it starts. I checked it holds even at `5e-324` where the product itself underflows to zero.

I resolved the CHANGELOG conflict against main and pushed it to your branch rather than making you rebase, since main had moved a fair way.

Vedangalle:
Thanks @jeqcho - I appreciate the clear rationale and the merge. Glad the API compatibility, validation ordering, and underflow handling proved useful. I’ll keep an eye on the downstream KitchenBench integration and any follow-up gaps that surface.

PR #162, opened 2026-07-22
feat(task): support seconds-based task horizons resolved against control rate (#160)

### LINKED ISSUE #160: [design] Seconds-based task horizons resolved against the embodiment's control rate at eval() time

Adityakk9031:
## Summary

Resolves #160 (and #161). 

`Task.max_steps` was previously a fixed integer step budget set at task construction time. However, benchmark tasks defined in physical time (e.g. KitchenBench) allot per-task seconds (`max_seconds`). Because task factories behind string registries are zero-arg and `Task` is embodiment-agnostic, tasks had no way to resolve physical durations into step counts prior to pairing with an embodiment.

This PR adds support for declaring task horizons in physical seconds (`Task(max_seconds=...)`), which `eval()` resolves into integer step budgets (`max_steps = math.ceil(max_seconds * control_hz)`) after pairing with an embodiment.

---

## Key Changes

- **`inspect_robots.task.Task`**: Made `max_steps: int | None = None` optional and added `max_seconds: float | None = None`. Added `__post_init__` validation enforcing mutual exclusion (exactly one of `max_steps` or `max_seconds` must be set) and positive numeric ranges. `task.envelope` raises `ConfigError` if called on an unresolved `max_seconds` task.
- **`inspect_robots.compat`**: Added a preflight compatibility check (`control_hz_missing` error issue) ensuring `embodiment.info.control_hz` is present and $> 0$ whenever `task.max_seconds` is specified.
- **`inspect_robots.eval`**: Computes `max_steps = math.ceil(task.max_seconds * control_hz)` at `eval()` time, passing the resolved integer budget to `bind_task` hooks and `rollout()`.
- **`inspect_robots.log`**: Added `max_seconds: float | None = None` to `EvalSpec` for reproducible log schema persistence.
- **`tests/test_task_horizon.py`**: Added unit tests covering task horizon validation, preflight `control_hz` compatibility checks, fractional `math.ceil` resolution, and `EvalLog` persistence.

---

## Verification

- **Unit Tests**: Executed full test suite (`pytest --cov`). All **716 tests passed**.
- **Coverage**: **100% statement and branch coverage** maintained across all core modules.
- **Linting & Types**: Passed `ruff check .`, `ruff format --check .`, and `mypy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,37 +39,13 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 
-- **Policy lifecycle hook: `on_trial_end`** — policies can now hook into
-  the end of a trial to persist state or artifacts. The orchestrator calls
-  `policy.on_trial_end(record, log_dir, run_id)` and any metadata the policy
-  attaches to `record.metadata` is persisted in the final `EvalLog`. Hook
-  failures are caught and logged as trial errors, preventing them from
-  crashing the overall evaluation (#40).
-- **Agent plugin transcript persistence** — `LLMAgentPolicy` now implements
-  `on_trial_end` to persist its full conversation transcript (tool calls,
-  observations, system prompts) to a JSONL file per trial under
-  `<log-dir>/transcripts/<run_id>/<scene_id>-e<epoch>.jsonl`. Camera images
-  are stripped from the transcript to save space, as they are already
-  recorded in the frame store. The relative path to the transcript is
-  stored in the trial's metadata for easy post-hoc analysis (#40).
-- **Agent plugin:** `-P wire=responses` selects the OpenAI Responses API wire,
-  so reasoning effort works together with function tools on recent OpenAI
-  models (Chat Completions rejects the combination, observed on
-  `gpt-5.6-sol`). The chat-wire rejection now names the fix in its error
-  message (#131).
-- **`inspect-robots view LOG.json`**: render a saved eval log as a
-  self-contained HTML report with run metadata, scores, scene results,
-  collapsible policy conversations, highlighted agent notes, and the camera
-  frames the model saw in `--store-frames` runs. `--no-frames` keeps
-  placeholders and `--frames-budget` controls the inline payload limit (#132,
-  #141).
-- **`inspect-robots eval-set TASK [TASK ...]`**: run several registered tasks
-  against one resolved policy/embodiment pair in a single invocation, matching
-  task names exactly or by shell-quoted `fnmatch` glob (e.g.
-  `'kitchenbench/*'`). Thin CLI wrapper over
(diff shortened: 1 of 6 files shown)


Adityakk9031:
@jeqcho have a look

Adityakk9031:
@jeqcho all check pass

jeqcho:
@Adityakk9031 closing this in favour of #161, reasoning in this comment (https://github.com/robocurve/inspect-robots/pull/161#issuecomment-5087618009).

I would love to see a follow-up PR from you. The bool check is the piece I want kept from #162. #161 validates max_steps >= 1 but doesn't reject True, so Task(max_steps=True) silently becomes a horizon of 1. Yours catches it. Open that as its own PR once #161 lands, otherwise it'll conflict, and I'll take it.

Plenty else open across github.com/robocurve if you want something in the meantime.

Adityakk9031:
ok

PR #181, opened 2026-07-27
fix(agent): raise ConfigError instead of ValueError for constructor v…

### LINKED ISSUE #168: Agent plugin: constructor validation errors are half ValueError, half ConfigError

Atif0110:
…alidation

Converts the remaining 6 constructor validation checks in LLMAgentPolicy (wire, speed, effort, max_llm_calls, max_speed_frac, max_output_tokens) from ValueError to ConfigError, consistent with the other checks in the same constructor and with the CLI's guided-error rendering.

Fixes #168

## Summary

What does this change do and why?

## Changes

-

## Checklist

- [x] `mypy` passes (strict)
- [ ] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test

## Related

Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ All notable changes to this project are documented here. The format is based on
 
 ### Changed
 
+- **Agent plugin:** `LLMAgentPolicy` constructor validation for `wire`, `speed`,
+  `effort`, `max_llm_calls`, `max_speed_frac`, and `max_output_tokens` now
+  raises `ConfigError` instead of `ValueError`, consistent with the other
+  constructor checks and the CLI's guided-error rendering (#168).
+
 - **Agent plugin:** runtime camera dropouts in `images=on_demand` now reject
   `take_pic` without treating a well-formed call as a tool error, so a single
   dropout no longer errors the trial. The first world-state rejection in one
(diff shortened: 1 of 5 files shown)


jeqcho:
Thanks for the contribution, @Atif0110! This branch currently has merge conflicts with `main`, so CI wasn't able to run. Could you rebase onto the latest `main` and resolve the conflicts? We'll review as soon as it's green. Thank you! 🙏

Atif0110:
Hey Jay, went to rebase this and found the same fix already landed on main in the meantime, identical ConfigError conversion for the same 6 checks. Since my change would be a no op now, I'll close this out. Thanks for the quick review!

Issue #173, opened 2026-07-26
agent: on-demand camera capture and chained tool calls


jeqcho:
Two related gaps in the agent policy's perception loop.

**1. Images are force-fed every turn.** `_observation_content` attaches every camera frame to every observation message. On a long trial that is the dominant token cost, and the model has no way to say "I do not need to look right now". There should be a mode where frames arrive only when the model asks for them, via a `take_pic` tool.

**2. Tool calls cannot be chained.** `act()` executes the first tool call of an assistant turn and answers every extra with `ignored: one tool call per turn`. The natural pattern for on-demand vision is "move, then look at the result", which today costs a full extra LLM round-trip after the motion has already played out.

Chaining a capture after a motion has to respect playout: the frames must come from the observation delivered *after* the whole speed-limited chunk has finished, not from the pre-motion observation, and the model should be told when a motion was cut short (early replanning, termination) instead of being handed frames labelled as post-arrival.

PR #182, opened 2026-07-27
setup: offer the agent's on-demand camera mode, defaulting to on_demand (#179)

### LINKED ISSUE #179: setup: offer the agent's on-demand camera mode, defaulting to on_demand
`-P images=on_demand` (#173, #175) lets the agent call `take_pic` for a frame instead of having every camera attached to every observation. It is the cheaper mode and the one a real rig usually wants, but nothing surfaces it: `images` defaults to `always` (`plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py:137`) and you only find the alternative by reading the plugin README.

`inspect-robots setup` is where a rig operator makes exactly this kind of choice, so the prompt belongs there, suggesting `on_demand` while leaving the library default alone.

## Why the wizard rather than flipping the default

Flipping `images` in the plugin would change behaviour for programmatic and CI callers who never run the wizard, and there is no performance data yet saying `on_demand` is better rather than just cheaper. Steering the humans who run setup gets the benefit where it matters without a breaking change, and the mode is stamped into `EvalSpec.policy_config` either way (`eval.py:291` does `asdict(policy.config)`), so any log says which mode produced it.

## Shape

The config format already supports this. `[policy.args]` is parsed at `_defaults.py:161` and scoped to the chosen policy via `policy_args_owner`, so the args do not leak if the policy changes later. Writing this by hand works today:

```ini
[defaults]
policy = agent

[policy.args]
images = on_demand
```

What is missing is the wizard.

- A prompt in `_setup.py`, gated on the selected policy being `agent`, defaulting to `on_demand`. The wording should state the tradeoff rather than just the option: the model calls `take_pic` when it wants a frame, which cuts tokens, but it has to remember to look.
- The renderer needs to emit `[policy.args]`. It currently manages only `embodiment.args` (see `_preserve_managed_args` at `_setup.py:577` and the render path around `:895`) and carries every other section through raw. Teaching it a second managed section is the fiddly part of this issue and the place bugs are most likely.
- Re-running setup should preserve an existing `images` value the way camera keys are preserved today.

## Docs

Two lines in `plugins/inspect-robots-agent/README.md` (77 and 151) say the default is `always`. Both should also say that `inspect-robots setup` suggests `on_demand`, so that someone reading the docs and then debugging a set-up machine is not surprised. `docs/guide/cli.md` and `docs/guide/quickstart.md` describe the wizard and want a line about the new prompt.

## Open question

Whether `on_demand` should simply become the plugin default is worth revisiting once there is data on how the two modes actually score. This issue deliberately does not settle that.

Adityakk9031:
Closes #179.

### Problem
`-P images=on_demand` (#173, #175) allows the `agent` policy to issue `take_pic` tool calls for frames on-demand instead of attaching all camera frames to every observation. While this is cheaper and preferable for real rigs, it was only discoverable via the plugin README. `inspect-robots setup` is where operators configure defaults, but it previously did not prompt for `agent` policy options.

### Solution
1. **Setup Wizard (`_setup.py`)**:
   - Added `_valid_images_mode()` validator (`"on_demand"` / `"always"`).
   - Gated a prompt on `configured_policy == "agent"` that suggests `on_demand` (explaining the token savings vs model responsibility tradeoff) while preserving any existing `images` configuration in `[policy.args]` on setup re-runs.
   - Updated `_render_config()` to render managed `[policy.args]` blocks alongside `[defaults]` and `[embodiment.args]` while preserving any unmanaged policy args.
2. **Documentation**:
   - Updated `plugins/inspect-robots-agent/README.md` to note that `inspect-robots setup` suggests `on_demand` while preserving the library default `always`.
   - Updated `docs/guide/cli.md` and `docs/guide/quickstart.md` wizard descriptions.
3. **Tests (`tests/test_setup.py`)**:
   - Added unit test `test_render_config_renders_policy_args`.
   - Added integration tests for prompt defaults (`on_demand`), explicit selection (`always`), and config preservation on setup re-runs.

### Verification
- `ruff check .` — All checks passed.
- `ruff format --check .` — All 108 files formatted.
- `pytest tests/test_setup.py` — **105 passed, 16 skipped** (POSIX/symlink-gated tests skipped on Windows as expected).


Adityakk9031:
@jeqcho have a look

jeqcho:
Thanks @Adityakk9031! CI on this branch is green, but it now has merge conflicts with `main`. Could you rebase and resolve? We'll review right after. Thank you! 🙏

Adityakk9031:
@jeqcho hey conflict are  fixed have a look

jeqcho:
Thanks @Adityakk9031 — you did resolve the conflicts, but `main` has moved again since your push on Aug 1 and the branch is now conflicting once more (sorry about the moving target! 🙏). Could you rebase one more time? We'll get a review in promptly after that so it doesn't drift again.

jeqcho (approved):
Thanks @Adityakk9031 — this lands exactly what #179 asked for, and the fiddly part (teaching the renderer a second managed section) came out clean.

What I checked:

- **Scope matches the issue.** The wizard suggests `on_demand`; the plugin default stays `always`, so programmatic and CI callers are untouched. The `policy_args_owner` scoping (`defaults.py:219`, `cli.py:1420`) keeps a carried `images` key inert if the policy later changes, so the raw carry-through in the renderer is safe.
- **Renderer.** The new `[policy.args]` block in `src/inspect_robots/_setup.py` (`_render_config`) mirrors the `[embodiment.args]` block exactly: managed keys first, unmanaged keys carried raw, multiline `\n` → `\n\t` escaping, empty-section suppression, and the section correctly excluded from the generic carried loop. No scoping hazards in `run_setup` — `policy_args` and `managed_policy_args` are always bound before `_render_config`, and every abort path returns before rendering.
- **Tests.** All new branches are covered (default suggestion, explicit `always`, preservation on re-run, invalid existing value falling back to `on_demand`, renderer unit test, empty-managed-args edge), which should keep the 100% coverage gate green.
- **Docs.** README, `cli.md`, and `quickstart.md` all updated per the issue.

Two tiny non-blocking nits:

- `plugins/inspect-robots-agent/README.md` line 222: the amended parenthetical runs long compared to the surrounding ~80-col wrap; consider rewrapping.
- When an existing config has an invalid `images` value, the wizard silently drops it and suggests `on_demand` (tested, and a reasonable choice) — a one-line dim note telling the operator their old value was discarded would be a nice touch, but happy to leave it.

**One ask before we can merge:** the branch is based on an older `main` and currently conflicts — `_setup.py` has moved a fair bit since (the wizard grew several sections). Could you rebase onto the latest `main`? Worth re-running the setup tests after the rebase since the scripted-input sequences assume the current six `[defaults]` prompts (they do still match today's suggestions). CI can gate it once the rebase is up, and I'm happy to merge then.

PR #209, opened 2026-07-28
fix(rollout): persist post-action frames

Vedangalle:
## Summary

Persist every distinct rollout observation through `FrameStore`, including the terminal post-action state.

Previously, the rollout stored only each step's pre-action observation. That left full camera arrays inside `StepResult.observation` and gave the terminal observation no sidecar reference, because it never becomes a later step's pre-action input.

## Changes

- store the reset observation at index `0` and each step result at `t + 1`;
- add `StepRecord.result_image_refs` for post-action frame handles;
- strip camera arrays from both recorded pre-action and post-action observations;
- reuse each stored post-action view on the next step, preventing duplicate writes;
- document the storage and compatibility contract in `plans/0032-post-action-frame-persistence.md`;
- add regression coverage for contiguous indexing, exact write count, and terminal-frame recovery.

A one-camera rollout with `n` steps now writes exactly `n + 1` frame files.

## Compatibility

- `result_image_refs` is appended with a default of `None`.
- Runs without `FrameStore` preserve the original observation and result objects.
- `StepRecord` is not part of the schema-versioned `EvalLog`, so no log migration is required.
- Core remains NumPy-only and no exported symbol changes are required.

## Verification

- [x] Tests added/updated, including terminal-state and exact-write-count coverage
- [x] Full upstream CI matrix passes on current `main`
- [x] Coverage: **100% line and branch**
- [x] `ruff check .` passes
- [x] `ruff format --check .` passes
- [x] Strict `mypy` passes
- [x] Core remains NumPy-only

The gap surfaced while building a process-aware evaluator on top of Inspect Robots: https://github.com/Vedangalle/inspect-robots-verifier

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 
+- `FrameStore` now persists each post-action observation once and exposes it
+  through `StepRecord.result_image_refs`. Stored records strip camera arrays
+  from both pre-action and post-action observations, and the terminal visual
+  state is recoverable for offline scoring.
+
 - The Rerun sink now sends a per-trial blueprint that groups labeled action
   dimensions by arm, overlays aligned measured state, and lays out cameras,
   transcript, events, and reward alongside the joint plots
--- a/plans/0032-post-action-frame-persistence.md
+++ b/plans/0032-post-action-frame-persistence.md
@@ -1,0 +1,69 @@
+# Post-action frame persistence
+
+## Problem
+
+`FrameStore` currently writes each step's pre-action `Observation`, then stores
+the unmodified `StepResult` in `StepRecord`. This has two consequences:
+
+1. `StepResult.observation.images` retains full camera arrays in memory, so a
+   stored trajectory does not satisfy the frame-store memory bound.
+2. The post-action observation from the final step has no sidecar reference.
+   Offline scorers therefore cannot reconstruct the terminal visual state.
+
+For non-terminal steps, the same post-action observation becomes the next
+pre-action observation and is stored one iteration later. The terminal
+observation has no next iteration and is lost from the sidecar sequence.
+
+## Contract
+
+With a `FrameStore`, every distinct rollout observation is written exactly once:
+
+- reset observation at frame index `0`;
+- result of step `t` at frame index `t + 1`.
+
+`StepRecord.observation` and `StepRecord.result.observation` contain no image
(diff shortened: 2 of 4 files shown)


jeqcho:
Heads up: main's history was rewritten today to purge 3 GB of accidentally committed eval logs (#212). Your branch was based on one of the rewritten commits, so I rebased it onto the new main and force-pushed (maintainer edits are enabled on this PR). No changes to your commits themselves; CI will re-run. If you have local work on this branch, run git fetch and rebase onto the updated remote branch before pushing.

Vedangalle:
@jeqcho When you have a chance, I’d appreciate your (any) feedback on the FrameStore contract here - particularly whether `result_image_refs` is the right interface for post-action and terminal observations. Happy to adjust quickly! :)

jeqcho (approved):
Thanks for this, @Vedangalle — and for the unusually thorough write-up. This closes a real gap: the terminal observation genuinely had no sidecar reference, and keeping full camera arrays inside `StepRecord.result` quietly defeated the frame-store memory bound (R5). I especially like that the indexing scheme is purely additive — the post-action frame of step `t` lands at `t + 1`, which is exactly where the next step's pre-action frame already lived, so existing on-disk layouts, the video exporter, and the viewer all keep working while gaining the terminal frame. The stored-view reuse to avoid duplicate writes is elegant, the identity-preserving no-`FrameStore` path is the right call, and the exact-write-count and terminal-recovery tests pin the contract down nicely.

A few small, non-blocking points — merging now, but a small follow-up PR for these would be very welcome:

1. **CHANGELOG placement:** with a `FrameStore` active, `step.result.observation.images` is now empty — any external scorer that read terminal images from the in-memory result must switch to `result_image_refs`. Since that's a behavior change (the intended fix, but still a change), could you move or duplicate the entry under a "Changed" heading and name the migration path explicitly?
2. **Docs touch-up:** `docs/guide/logging-and-rerun.md` describes `image_refs`/`FrameRef.path` as the authoritative mapping — worth a sentence introducing `result_image_refs` and the new `n + 1` frame count. Same for the small "frames are stored once per rollout step" note in `_video.py`'s `default_fps` docstring, which is now off by one in spirit.
3. **Tiny observation, no action needed:** the reset frame is now written before the first `next_action` call, so a trial that fails at step 0 leaves reset frames on disk where previously it left none. I think that's a feature (forensics), just flagging it as intentional.

To answer your direct question: yes, `result_image_refs` as a sibling of `image_refs` is the right interface — it keeps the pre/post pairing explicit per step rather than forcing consumers to reconstruct it from indices.

PR #217, opened 2026-07-29
feat(rollout): surface approver interventions to policy via observati…

### LINKED ISSUE #187: Surface approver interventions to the policy via observation.extra
## Current behavior

When an approver modifies an action, the rollout records an `approval_event(t, modified=True, detail="clamped")` in the trial event stream (rollout.py). That is the only place the intervention is visible. The policy never learns that its action was rewritten: the next observation shows the resulting state, but not why the commanded motion under-delivered.

For the `agent` policy this matters in practice. The system prompt pre-warns the model that approvers clamp out-of-bounds and too-fast actions, but after a clamp the model has to infer what happened from proprioceptive drift alone. It can misattribute the undershoot (controller loss, stiction, a bad target estimate) and spend LLM-call budget re-issuing the same out-of-bounds command. In `images=on_demand` mode the queued-capture narration reports the residual offset from target, which points at the affected joint but still not at the cause.

## Proposal

Stamp approver interventions from the steps since the last `act()` into `observation.extra` (e.g. `extra["approvals"]`: list of `{t, detail}` mirroring the logged approval events). Then:

- `inspect-robots-agent`: render them in `_observation_content()` as a line in the state text, e.g. `approver: 3 of 20 steps clamped (delta_clamped)`, so the model can react (choose a nearer target or smaller motion) instead of retrying blind.
- Other policies get the same signal for free; policies that ignore `extra` are unaffected.

The event stream in the EvalLog stays the source of truth; this is a read-only mirror of events already being recorded.

## Considerations

- Which approver metadata to expose: the `detail` flags (`clamped`, `delta_clamped`) are enough for the model to act on; per-dimension magnitudes could ride in later if needed.
- `SafetyAbort` needs no surfacing (the trial halts).
- Core stays NumPy-only; this is a plain dict entry.
- Needs a decision on whether `extra` is the right channel vs. a first-class field on `Observation` (plan 0001 binding resolutions may constrain this).

Adityakk9031:
### Description

Fixes #187 by surfacing safety approver interventions (such as `clamped` and `delta_clamped` rewrites) directly to policies in `observation.extra["approvals"]`.

Previously, when an approver rewritten an action, the modification was recorded into the trial's event stream in `EvalLog`, but never surfaced to `policy.act()`. This meant LLM/VLA policies had to infer clamps from proprioceptive drift alone, often misattributing undershoot to dynamics/stiction and re-issuing the same out-of-bounds action.

### Proposed Changes

- **Core (`inspect_robots`)**:
  - In `rollout.py`, track approval events in trial `store` under `_APPROVALS_KEY`.
  - Pass the accumulated list of approval intervention records (`{"t": t, "detail": detail}`) into `obs.extra["approvals"]` for every step when calling `controller.next_action(...)`.
  - Added unit test `test_rollout_surfaces_approvals_in_observation_extra` in `tests/test_rollout_hardening.py`.

- **Agent Policy (`inspect_robots-agent`)**:
  - In `_observation_content()`, render `observation.extra.get("approvals")` as a prompt text line (e.g. `approver: 3 step(s) modified (clamped, delta_clamped).`).
  - Allows LLM agent models to recognize safety gate interventions and adjust motion/targets rather than retrying blindly.
  - Added unit test `test_observation_content_renders_approver_interventions` in `plugins/inspect-robots-agent/tests/test_policy_e2e.py`.

### Verification

- Ran test suite for core: `pytest tests/` (861 passed).
- Verified agent policy observation content rendering tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ All notable changes to this project are documented here. The format is based on
   transcript, events, and reward alongside the joint plots
   ([plan 0041](plans/0041-rerun-arm-blueprint.md), #265).
 
+- **Approver interventions in observations:** safety approver rewrites (such as
+  clamping or delta-limiting) since the previous decision are now windowed into
+  `observation.extra["approvals"]`. `LLMAgentPolicy` renders them as aggregated prompt
+  lines (e.g. `approver: 3 step(s) modified (clamped ×3).`), enabling LLM/VLA models
+  to recognize safety interventions and adapt targets (#187, #217).
+
 - **Agent plugin (0.20.0):** `-P wire=gemini-live` now serves
   `google/gemini-robotics-er-2-streaming-preview` through Google's stateful
   v1beta Live API. The sync websocket client streams only new observations,
--- a/plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py
+++ b/plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py
@@ -16,6 +16,7 @@ import hashlib
 import json
 import os
 import sys
+from collections import Counter
 from collections.abc import Mapping
 from contextlib import suppress
 from dataclasses import dataclass
@@ -1000,7 +1001,16 @@ def _approvals_line(observation: Observation) -> str | None:
         return None
     total = len(approvals)
     details = [str(a.get("detail")) for a in approvals if isinstance(a, dict) and a.get("detail")]
-    detail_str = f" ({', '.join(details)})" if details else ""
+    if not details:
+        return f"approver: {total} step(s) modified."
+    counts = Counter(details)
+    formatted: list[str] = []
+    for flag, count in counts.items():
+        if count > 1:
+            formatted.append(f"{flag} \u00d7{count}")
+        else:
+            formatted.append(flag)
+    detail_str = f" ({', '.join(formatted)})"
     return f"approver: {total} step(s) modified{detail_str}."
 
 
--- a/plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py
+++ b/plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py
@@ -919,6 +919,16 @@ def _step_label(observation: Observation) -> str:
     return f"step {step}" if isinstance(step, int) else ""
 
 
+def _approvals_line(observation: Observation) -> str | None:
+    approvals = observation.extra.get("approvals")
+    if not isinstance(approvals, list) or not approvals:
+        return None
+    total = len(approvals)
+    details = [str(a.get("detail")) for a in approvals if isinstance(a, dict) and a.get("detail")]
+    detail_str = f" ({', '.join(details)})" if details else ""
+    return f"approver: {total} step(s) modified{detail_str}."
+
+
 def _observation_content(
     observation: Observation,
     state_labels: tuple[str, tuple[str, ...]] | None = None,
@@ -932,6 +942,9 @@ def _observation_content(
     if observation.instruction:
         lines.append(f"Instruction: {observation.instruction}")
     lines.extend(_state_lines(observation, state_labels))
+    app_line = _approvals_line(observation)
+    if app_line is not None:
+        lines.append(app_line)
     if narration is not None:
         lines.append(narration)
     parts: list[dict[str, Any]] = [{"type": "text", "text": "\n".join(lines)}]
--- a/plugins/inspect-robots-agent/tests/test_policy_e2e.py
+++ b/plugins/inspect-robots-agent/tests/test_policy_e2e.py
@@ -605,6 +605,22 @@ def test_transcript_is_none_before_first_reset() -> None:
     assert policy.transcript() is None
 
 
+def test_observation_content_renders_approver_interventions() -> None:
+    obs = Observation(
+        state={"q": np.array([0.0])},
+        extra={
+            "env_step": 2,
+            "approvals": [
+                {"t": 0, "detail": "clamped"},
+                {"t": 1, "detail": "delta_clamped"},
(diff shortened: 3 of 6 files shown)


Adityakk9031:

jeqcho (changes requested):
Thanks for taking this on — the mechanism is the right shape (rollout mirrors approval events into the store, snapshots them into `extra` before `next_action`, agent policy renders a line), the per-trial `store` created at `src/inspect_robots/rollout.py:192` means nothing leaks across trials/epochs, and both new tests are well targeted (nice touch asserting the first inference sees `[]` in `tests/test_rollout_hardening.py::test_rollout_surfaces_approvals_in_observation_extra`). Both touched test files pass locally (152 passed). A few things need addressing before merge:

**1. The rendered approvals line grows unboundedly — `plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py:922-929`**

`_approvals_line` joins one detail string per intervention with no aggregation. In exactly the scenario #187 describes — a policy repeatedly issuing out-of-bounds actions against a `ClampApprover` — a few hundred clamped steps produce `approver: 400 step(s) modified (clamped, clamped, clamped, …)` with hundreds of repeated tokens, injected into *every* subsequent observation. Please aggregate by flag, e.g. `collections.Counter` over the details, rendering something like `approver: 400 step(s) modified (clamped ×398, delta_clamped ×2).` — and extend `test_observation_content_renders_approver_interventions` to cover duplicate flags.

**2. Cumulative-since-reset semantics diverge from the issue and dilute the signal — `src/inspect_robots/rollout.py:220`**

#187 proposes surfacing interventions "from the steps since the last `act()`". The PR instead passes the full trial history on every step. Beyond size, this weakens the actionable signal: what the model needs to know is "was my *latest* motion clamped?", and under cumulative semantics that is drowned out once early interventions accumulate (the line eventually reads the same whether or not the most recent chunk was touched). The rollout already tracks inference boundaries via `prev_inferences` / `_INFER_KEY` (lines 219, 232), so windowing is cheap — e.g. record the approvals-list length at each inference and pass only the tail since the previous one. If you have a deliberate reason to prefer cumulative (`t` fields do let a sophisticated policy diff), please say so on the issue and document the chosen semantics; but for the agent-policy rendering, "since your last decision" is the more useful contract.

**3. Missing CHANGELOG entry**

CONTRIBUTING.md (step 4) asks for a `CHANGELOG.md` entry under "Unreleased". This is a policy-visible contract addition, so it belongs under "Added", referencing #187/#217 in the style of the existing entries.

**4. Document the newly reserved `extra` key — `src/inspect_robots/types.py:32-35`**

The `Observation` docstring currently says the rollout "injects the current step into the policy-facing observation's `extra["env_step"]` and reserves that key; embodiments should not set it." Since `rollout.py:223` now also unconditionally sets (and would silently overwrite) `extra["approvals"]`, please extend that docstring to reserve `"approvals"` the same way, including its shape (`list of {"t": int, "detail": str | None}`) and the fact that it is always present, possibly empty.

**Minor (optional):**

- `approvals = list(store.get(_APPROVALS_KEY, []))` at `rollout.py:220` shallow-copies the list, but the entry dicts are shared by reference across every observation for the rest of the trial (and with the store). A policy that mutates an entry would corrupt what later inferences see. Copying the dicts (or storing immutable tuples) would harden this cheaply.
- `_approvals_line`'s defensive `isinstance` checks are welcome given `extra` is an open channel; no change needed there.

Nothing here is structural — the core wiring is sound and matches the issue's intent. Happy to re-review once the rendering aggregation, windowing decision, changelog, and docstring are in. Thanks again for the contribution! 🙏

Adityakk9031:
Thanks @jeqcho for the thorough review! I've updated the PR to address all points:

1. **Aggregated Prompt Rendering**: `_approvals_line()` now uses `collections.Counter` to group duplicate detail flags into count strings (e.g., `approver: 400 step(s) modified (clamped ×398, delta_clamped ×2).`). Added test coverage for duplicate flags in `test_policy_e2e.py`.
2. **Windowed Semantics**: `rollout.py` now tracks `_rollout_last_approvals_idx` across inferences and passes only the intervention records since the previous `act()` call.
3. **Mutation Hardening**: Each entry in `obs.extra["approvals"]` is dictionary-copied so policy mutations cannot pollute the rollout store or future observations.
4. **Documentation & Changelog**: Updated the `Observation` docstring in `types.py` reserving `extra["approvals"]` and added a `CHANGELOG.md` entry under "Unreleased" -> "Added".

jeqcho (approved):
Thank you for the quick and thorough turnaround — this is exactly what the review asked for, and then some. The Counter-based aggregation reads great in the prompt, the windowing rides the existing `_INFER_KEY` inference boundaries just as hoped (and I verified the append ordering puts each step's intervention in the correct window), the docstring and CHANGELOG both document the new contract, and the extended tests are genuinely sharp — I especially appreciated the mutation probe in `test_rollout_surfaces_approvals_in_observation_extra` proving a misbehaving policy can't corrupt later windows. Both touched test files pass locally and CI is fully green, so this is good to merge. 🎉

Two tiny optional nits, purely take-it-or-leave-it (fine as a follow-up or never):

1. `"_rollout_last_approvals_idx"` appears as a raw string twice in `rollout.py` — hoisting it to a module constant next to `_APPROVALS_KEY` would keep the store keys consistent.
2. The third-window assertions in the new rollout test are guarded by `if len(policy.captured_extras) > 2:`, so they'd silently skip if the trial ever shortened; the unconditional first/second-window asserts already carry the contract, but dropping the guard (or asserting the length) would make the test fully strict.

Thanks again for contributing this — it closes a real gap for LLM/VLA policies fighting the safety gate, and the follow-through on every review point made this a pleasure to re-review. 🙏 Merging!

PR #228, opened 2026-07-30
refactor(cli): share the --epochs guard and drop the doubled message

### LINKED ISSUE #152: refactor(cli): extract epochs ConfigError handling into a shared helper and polish the message
Follow-up to #145, which was fixed by #146. Two PRs raced on that issue and each got something right; this issue collects what the merged one left on the table.

Current state after #146: `_cmd_run` and `_cmd_eval_set` each carry an inline `try/except ConfigError` around their `replace(task, epochs=...)` call, with the `ConfigError` import repeated mid-function in both.

Wanted:

- Extract the two blocks into a shared `_apply_epochs_or_exit` helper next to `_resolve_or_exit`, following the same `raise SystemExit(str(exc)) from exc` idiom. #148 already had this shape, including moving `replace` into the helper and importing `Task` under `TYPE_CHECKING`; it is the reference.
- Polish the message while at it: `--epochs: Epochs count must be >= 1, got 0` doubles up. The issue proposed `--epochs must be >= 1, got 0`. Keep the task attribution in the eval-set path.
- Consider `pytest.raises(..., match=...)` exact-message pinning in the epochs tests (again per #148) rather than substring asserts.

Small and self-contained; a good second contribution for anyone who worked on #145.

ptlnextdoor:
Closes #152. Claiming this one per your note on #148 — thanks for opening it, and no worries at all about the race on #145.

Rebased onto current `main`, so this is on top of #146 rather than a resubmit of #148.

## What changed

**Shared helper.** `_cmd_run` and `_cmd_eval_set` each carried an inline `try/except ConfigError` around `replace(task, epochs=...)`, with the `ConfigError` import repeated mid-function. Both now call `_apply_epochs_or_exit`, which sits next to `_resolve_or_exit` and follows the same `raise SystemExit(...) from exc` idiom. `replace` moved into the helper and `Task` is under `TYPE_CHECKING`, so neither call site needs a local import.

The `eval-set` loop collapses back to a comprehension, since the `patched` list only existed to hold the try/except:

```python
tasks = [_apply_epochs_or_exit(t, args.epochs, attribute_task=True) for t in tasks]
```

**The doubled message.** Was `--epochs: Epochs count must be >= 1, got 0`; now:

```
--epochs must be >= 1, got 0
--epochs (task 'cubepick-reach') must be >= 1, got -1
```

Task attribution is kept on the `eval-set` path via `attribute_task=True` — `run` has one task, `eval-set` needs to say which of several rejected the flag.

One note on why the message can be phrased in the flag's own terms rather than wrapping `exc`: `Task.__post_init__` re-validates every field, but the task was already valid and only `epochs` changed, so the epoch-count check is the only rejection this call can newly trigger. The other `ConfigError`s in `task.py` are the `max_steps`/`max_seconds` invariants (already satisfied) and two in `envelope()`, which is a different code path. I did write a defensive fallback for a non-epochs `ConfigError` first, then deleted it — it was unreachable, and it was the one line keeping coverage at 99.96%. Flagging in case you'd rather have the guard and a `# pragma: no cover`.

**Tests.** Both messages are now pinned with `==` instead of substring asserts, so a regression to the doubled wording fails rather than passing.

## Verification

- `uv run ruff check .` + `ruff format --check .` → clean
- `uv run pytest --cov` → **920 passed, 100% coverage** (line + branch)
- Manually confirmed both messages:
  ```
  $ inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --epochs 0
  --epochs must be >= 1, got 0
  $ inspect-robots eval-set cubepick-reach --policy scripted --embodiment cubepick --epochs -1
  --epochs (task 'cubepick-reach') must be >= 1, got -1
  ```

`uv run mypy` reports 2 errors in `scripts/gen_api_docs.py`. Those are pre-existing — identical with these changes stashed — so I left them alone.


jeqcho (commented):
Thanks for coming back for this one — and for a PR body that answers the reviewer's questions before they're asked. This lands #152 exactly as scoped, and it reads like it grew here: helper next to `_resolve_or_exit`, same `from exc` idiom, `replace` and `ConfigError` pulled into the helper so neither call site needs a mid-function import.

Things I particularly liked:

- The `attribute_task=True` keyword keeps message construction in one place while letting the `eval-set` loop collapse back to a comprehension — that's the shape the issue was hoping for.
- Your unreachability argument for the non-epochs `ConfigError` checks out against `task.py`: the task was already valid and only `epochs` changed, so the epoch-count check is the only rejection `replace` can newly trigger. Deleting the defensive fallback was the right call — an unreachable branch with a `# pragma: no cover` documents a fear, not a behavior, and the comment you left in the `except` explains the reasoning where the next reader needs it.
- Pinning both messages with `==` instead of substring asserts means a regression to the doubled wording actually fails. I also checked `test_cli_run_closes_embodiment_when_validation_raises` — its `"--epochs" in` assert stays green with the new wording, so the close-on-failure invariant is undisturbed.

One small thing before merge:

1. `CHANGELOG.md` — the contributing guide asks for an entry under "Unreleased", and the existing #145 entry there still describes the pre-refactor shape ("surface it through the existing `_resolve_or_exit` pattern"). Since that fix hasn't shipped yet, could you amend that entry in place rather than adding a new one? A sentence covering the shared `_apply_epochs_or_exit` helper and the final message wording (`--epochs must be >= 1, got N`, with `(task 'NAME')` on the `eval-set` path), plus a #152 reference, would do it.

Happy to do a final pass and merge once that's pushed — this is very close.

jeqcho:
Thank you @ptlnextdoor — this is exactly what #152 asked for, and then some. The `attribute_task` kwarg is the right call: it keeps the task attribution where it belongs and lets the `eval-set` loop collapse back to a comprehension, and you matched the `raise … from exc` idiom from `_resolve_or_exit` precisely. The exact `==` message pins in the tests are appreciated too — a regression to the doubled wording now fails loudly instead of slipping past a substring check. And thanks for the reachability note on the dropped defensive guard; I agree it's unreachable, and phrasing the message in the flag's own terms reads much better. No pragma needed.

Two small things before we can land it:

1. Recent merges have put this branch in conflict with `main` — could you rebase onto the latest `main`?
2. While you're in there, please add a short entry to `CHANGELOG.md` under **Unreleased** (per the PR checklist).

Once it's rebased and CI is green again this is good to go. Thanks for picking this up — great second contribution.

jeqcho (approved):

PR #250, opened 2026-08-01
fix(cli): reject --max-steps < 1 with a guided error instead of a traceback

### LINKED ISSUE #248: [bug] run: --max-steps < 1 exits with a raw ConfigError traceback instead of a clean parser error
**Describe the bug**

`inspect-robots run --instruction ... --max-steps 0` (or any value `< 1`) exits with a raw
`ConfigError` traceback instead of the clean one-line error the CLI uses everywhere else.

The neighbouring `--epochs` flag validates the exact same way and *is* handled cleanly, so this
looks like an oversight rather than a deliberate difference.

**To reproduce**

```bash
inspect-robots run --instruction "wipe the table" --policy scripted --embodiment cubepick --epochs 0 --no-prompt
# --epochs: Epochs count must be >= 1, got 0        <- clean, exit 1

inspect-robots run --instruction "wipe the table" --policy scripted --embodiment cubepick --max-steps 0 --no-prompt
# Traceback (most recent call last):                <- raw traceback, exit 1
```

Full traceback for the second command:

```
  File ".../bin/inspect-robots", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File ".../inspect_robots/cli.py", line 2168, in main
    return _cmd_run(args)
  File ".../inspect_robots/cli.py", line 1254, in _cmd_run
    task = Task(
        name="adhoc",
    ...<3 lines>...
        metadata={"instruction": args.instruction, "adhoc": True},
    )
  File "<string>", line 10, in __init__
  File ".../inspect_robots/task.py", line 79, in __post_init__
    raise ConfigError(f"Task {self.name!r}: max_steps must be >= 1, got {self.max_steps}")
inspect_robots.errors.ConfigError: Task 'adhoc': max_steps must be >= 1, got 0
```

`--max-steps -5` behaves the same way.

**Expected behavior**

Same shape as `--epochs`:

```
--max-steps: max_steps must be >= 1, got 0
```

with a non-zero exit and no traceback.

**Additional context**

The ad-hoc `Task(...)` construction at `cli.py:1252-1260` is not wrapped, while the `--epochs`
path immediately below at `cli.py:1267-1273` catches `ConfigError` and calls `parser.error(...)`.
`main()` (`cli.py:2168`) has no top-level handler, so the exception reaches the user.

Happy to send a PR mirroring the `--epochs` handling if you'd like it.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

Rishet11:
Fixes #248.

`inspect-robots run --instruction ... --max-steps 0` (or any value `< 1`) reached
`Task.__post_init__` and surfaced the resulting `ConfigError` as an uncaught
traceback, because `main()` has no top-level handler.

```
before:  Traceback (most recent call last): ... ConfigError: Task 'adhoc': max_steps must be >= 1, got 0
after:   --max-steps must be an integer >= 1, got 0
```

**Approach.** I range-check the flag up front rather than catching `ConfigError`
around the `Task(...)` construction. Two reasons:

- Wrapping the constructor would cover more than the bug needs (it also spans
  scorer resolution and `Scene(...)`), so a future edit to that call site could
  get mislabelled as a `--max-steps` error.
- It keeps the message single-prefixed. Catching the `ConfigError` yields
  `--max-steps: Task 'adhoc': max_steps must be >= 1, got 0`, since the task name
  is already baked into that message.

This matches how the other numeric flags in the module are validated
(`--fps must be a positive finite number`, `--frames-budget must be a
non-negative finite number`).

The config-file source of `max_steps` is already validated in `defaults.py`, so
the CLI flag was the only unchecked path into that `ConfigError`.

**Tests.** Parametrized over `0` and `-5` in `tests/test_registry_cli.py`,
following the existing `test_cli_run_zero_epochs_exits_with_guided_error` style.
Verified they fail with the raw traceback before the change.

Local: `pytest` 992 passed / 3 skipped, coverage 100%, `ruff check` clean,


Rishet11:
Noticed #152 and #228 after opening this, and adjusted to match: the message now speaks in the flag's own terms (`--max-steps must be >= 1, got 0`) rather than echoing `Task`'s internal wording, and the test pins the exact string instead of substring-asserting.

No overlap with #228 in the diff. That PR refactors the `--epochs` guard around `cli.py:590`; this only adds a range check next to the existing `--max-steps` conflict checks, and deliberately leaves the epochs block untouched. Happy to rebase if #228 lands first.

jeqcho (approved):
Thanks @Rishet11 — this is a model bug report and fix pair. The write-up in #248 made verification trivial, and the PR body's reasoning for range-checking the flag up front instead of wrapping the `Task(...)` construction is exactly right: it keeps the error single-prefixed and avoids mislabelling future failures from scorer resolution or `Scene(...)`.

A few things I checked and liked:

1. The check sits after the `--task`/`--instruction` cross-checks, so `--task X --max-steps 0` still gets the more specific "only applies to --instruction runs" message — correct precedence.
2. Placing it in `_cmd_run` rather than `_check_shared_run_conflicts` is the right call, since `--max-steps` only exists on the `run` subparser and the shared helper also serves `eval-set`.
3. The message shape matches our other numeric-flag errors (`--max-action-delta`, `--frames-budget`, `--fps`), and pinning the exact message in the parametrized tests guards against regressing to the double-prefixed `Task 'adhoc'` wording.
4. Config-file and `config set` paths were already validated in `defaults.py`, so this closes the last route into that `ConfigError` — confirmed.

No changes requested — merging shortly. Much appreciated!

jeqcho:

Rishet11:

jeqcho (approved):

PR #251, opened 2026-08-01
fix(cli): reject an out-of-range --port before binding the view server

### LINKED ISSUE #249: [bug] view --serve: out-of-range --port crashes with an uncaught OverflowError
**Describe the bug**

`inspect-robots view LOGS --serve --port N` crashes with a raw `OverflowError` traceback when `N`
is outside 0-65535. The bind is wrapped in a handler that catches `OSError` only, but an
out-of-range port raises `OverflowError`, which escapes.

**To reproduce**

```bash
inspect-robots run --task cubepick-reach --policy scripted --embodiment cubepick --no-prompt
inspect-robots view logs/ --serve --port 99999
```

```
  File ".../inspect_robots/cli.py", line 1797, in _serve_view_directory
    server = ThreadingHTTPServer((bind_host, port), handler)
  File ".../socketserver.py", line 457, in __init__
    self.server_bind()
  File ".../http/server.py", line 136, in server_bind
    socketserver.TCPServer.server_bind(self)
  File ".../socketserver.py", line 478, in server_bind
    self.socket.bind(self.server_address)
OverflowError: bind(): port must be 0-65535.
```

`--port -1` fails the same way.

**Expected behavior**

A clean parser error, consistent with the other `view` flag validation:

```
--port: must be between 0 and 65535, got 99999
```

An in-range but unavailable port (e.g. one already bound) is already handled well by the existing
`OSError` branch, so only the range case is missing.

**Additional context**

`cli.py:1796-1798` catches `OSError` around `ThreadingHTTPServer(...)`. Two options: widen that to
`(OSError, OverflowError)`, or validate the range up front next to the existing
"`--port` requires `--serve`" check, which would give a better message. I'd lean toward the
up-front check.

Happy to send a PR if you'd like it.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

Rishet11:
Fixes #249.

`inspect-robots view LOGS --serve --port N` crashed with an uncaught
`OverflowError` for any `N` outside 0-65535. The bind at
`_serve_view_directory` is wrapped in an `except OSError`, but
`ThreadingHTTPServer` raises `OverflowError` for an out-of-range port, so it
escaped as a traceback.

```
before:  OverflowError: bind(): port must be 0-65535.
after:   --port must be between 0 and 65535, got 99999
```

**Approach.** Range-check up front, next to the existing
`--port requires --serve` guard, rather than widening the `except` at the bind
site. That keeps the message specific about which flag is wrong, and matches the
surrounding validation style (`--frames-budget must be a non-negative finite
number`).

An in-range but *unavailable* port (already bound) is unaffected and keeps its
current `OSError` handling. `--port 0`, which asks the OS to auto-assign, still
works and is still covered by the existing serve tests.

**Tests.** Parametrized over `-1`, `65536`, `99999` in
`tests/test_registry_cli.py`, next to `test_view_serve_options_require_serve`.

Local: `pytest` 993 passed / 3 skipped, coverage 100%, `ruff check` clean,


Rishet11:
Housekeeping: I opened five small fixes at once, and all five originally added their CHANGELOG entry at the same spot, so every pair collided. I've re-anchored them to different points in the `### Fixed` list — 8 of the 10 pairs now merge cleanly in any order.

The one remaining overlap for this PR is with #250. Whichever you merge second will want a one-line CHANGELOG rebase; ping me and I'll push it immediately. No code overlap, only the changelog line.

jeqcho (approved):
Thanks @Rishet11 — another well-scoped fix, and the issue → PR write-up made this a very easy review.

What I checked:

- The bug is real: `--port` is `type=int` with no range constraint, and `_serve_view_directory` wraps the bind in `except OSError` only, so `ThreadingHTTPServer`'s `OverflowError` escapes as a traceback exactly as #249 describes.
- Validating up front next to the `--port requires --serve` guard was the right call over widening the `except` — the message names the flag, and the ordering means a bad port without `--serve` still gets the more fundamental "requires --serve" error first.
- Port semantics are correct at both edges: `0` (OS auto-assign) stays allowed, and the displayed URLs already use `server.server_address[1]`, so the auto-assigned port shows correctly; an in-range-but-busy port keeps its existing `OSError` path untouched.
- The message follows the house style of the neighboring guards (`--frames-budget must be a non-negative finite number`, `--max-action-delta must be finite and > 0, got ...`) — nice that you matched that rather than the `--port:` prefix form sketched in the issue.
- Tests cover both out-of-range directions plus the reproducer's value, and pinning the full message (with the comment explaining why) is a reasonable step up from the `match=` style next door given the whole point is which error surfaces.

CHANGELOG entry is accurate and in the right section. No changes requested — LGTM, will merge shortly.

jeqcho:

Rishet11:

jeqcho (approved):
Thanks @Rishet11 for another well-scoped fix, and for the thorough issue write-up in #249 to go with it!

I verified the bug and the fix:

- **Bug is real.** `--port` is parsed with a plain `type=int` (`src/inspect_robots/cli.py`, `p_view.add_argument("--port", ...)`), and the bind site in `_serve_view_directory` only catches `OSError` around `ThreadingHTTPServer((bind_host, port), handler)`, so an out-of-range port escapes as an uncaught `OverflowError` exactly as described.
- **Fix is correct and well-placed.** The range check sits with the existing `--port requires --serve` / `--host requires --serve` guards in `_cmd_view`, before the `8300` default is applied, and the `args.port is not None` guard keeps the default path untouched. The message style matches the neighboring post-parse validations (`--frames-budget must be a non-negative finite number`, `--max-action-delta must be finite and > 0, got ...`), including the `got {value}` suffix.
- **Edge cases considered.** Allowing `--port 0` (OS auto-assign) is the right call — the serve path already reports the actual bound port via `server.server_address`, and existing serve tests rely on it. Notably, the existing `_port_number` argparse helper (used elsewhere) requires 1–65535, so it couldn't be reused here without breaking `--port 0`; the explicit 0–65535 check is the correct range for `bind()`. An in-range-but-busy port correctly keeps its `OSError` handling.
- **Tests.** The parametrized test over `-1` / `65536` / `99999` pins the exact message, sits next to `test_view_serve_options_require_serve`, and reuses the same fixtures/log-setup pattern. CI is fully green, including the 100% coverage gate, mypy strict, and ruff.

One take-it-or-leave-it nit: the CHANGELOG hunk introduces a double blank line between the `thought_signature` entry and the new one (the entry adds a leading blank line where one already exists). Purely cosmetic — fine to fix in a follow-up or leave as is.

Great contribution — LGTM.

jeqcho (approved):

PR #256, opened 2026-08-01
fix(registry): set the discovery flags only after discovery succeeds

### LINKED ISSUE #255: [bug] registry: a failed builtins import leaves the registry permanently empty
**Describe the bug**

`registry._ensure_loaded()` sets `_loaded_builtins = True` *before* the
`import inspect_robots._builtins` that it guards. If that import raises, the flag stays set, so every
later call takes the "already loaded" path and proceeds with an **empty registry** instead of
retrying or re-raising.

The first call surfaces the real error. Every call after it silently reports that no components
exist, which turns a clear import failure into a confusing "no embodiment named 'cubepick';
available: []".

**To reproduce**

```python
import builtins, sys
import inspect_robots.registry as reg

# simulate a builtins module that fails to import (a syntax error in a local
# edit, a broken optional dependency, a half-installed environment)
reg._loaded_builtins = False
reg._loaded_entrypoints = True
reg._FACTORIES = {k: {} for k in reg._FACTORIES}
sys.modules.pop("inspect_robots._builtins", None)

real_import = builtins.__import__
def boom(name, *a, **k):
    if name == "inspect_robots._builtins":
        raise RuntimeError("simulated broken builtins import")
    return real_import(name, *a, **k)
builtins.__import__ = boom

try:
    reg._ensure_loaded()
except RuntimeError as e:
    print("1st call raised (expected):", e)

builtins.__import__ = real_import          # the problem is now gone
reg._ensure_loaded()                        # but the registry stays empty forever
print("registered embodiments:", sorted(reg._FACTORIES["embodiment"]))
print("_loaded_builtins:", reg._loaded_builtins)
```

Output:

```
1st call raised (expected): simulated broken builtins import
registered embodiments: []
_loaded_builtins: True
```

**Expected behavior**

The flag should only be set once the import has actually succeeded, so a later call either retries
or raises the same error again. Silently serving an empty registry is the worst of the three
options.

**Additional context**

`registry.py:79-85`. The entry-point loop below it has the same flag-before-work shape, but each
iteration is individually guarded by its own `try`, so a broken plugin there is already handled
correctly and warns by name. It's specifically the builtins import that can leave the module in a
permanently wrong state.

Fix is a two-line reorder (set each flag after its work completes).

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

Rishet11:
Fixes #255.

`_ensure_loaded` set `_loaded_builtins` / `_loaded_entrypoints` **before** the
import and scan they guard. If either raised, the flag stayed set, so every later
call took the "already loaded" path and served an **empty registry** instead of
retrying or re-raising.

The first call surfaces the real error; every call after it silently reports that
no components exist:

```
1st call raised (expected): simulated broken builtins import
registered embodiments: []
_loaded_builtins: True
```

Fix is to set each flag once its work completes.

The per-plugin `try/except` inside the entry-point loop is untouched — a single
broken plugin still warns by name and does not abort discovery. This only changes
what happens when the scan or the builtins import itself fails.

**Tests.** Two regression tests in `tests/test_registry_cli.py`, next to the
existing `test_entrypoint_discovery`. Both were confirmed to fail against the
unfixed code and pass with it.

Local: `pytest` 992 passed / 3 skipped, coverage 100%, `ruff check` clean,


jeqcho (approved):
Thanks @Rishet11 — this is a great catch, and the issue write-up with the repro made it easy to verify. Silently serving an empty registry after a transient import failure is exactly the kind of state bug that turns into a confusing "no embodiment named 'cubepick'; available: []" report far from the real cause, and moving each flag to after the work it guards is the right minimal fix.

A few things I checked and liked:

- The reorder is safe against re-entrancy: `_builtins` only runs registration decorators at import time, so nothing re-enters `_ensure_loaded()` mid-import.
- Retries are idempotent: a failed builtins import is evicted from `sys.modules` and re-registers cleanly, and the entry-point rescan keeps its first-wins `setdefault` semantics with no duplicates.
- The per-plugin warn-and-continue behavior is deliberately untouched, so one broken plugin still doesn't abort discovery.
- Both regression tests are well isolated (flags, `entry_points`, `__import__`, and the `sys.modules` deletion all restored via monkeypatch), and I appreciate that you confirmed they fail against the unfixed code.

Changelog entry matches house style and CI is green across the matrix. Merging — thanks for the contribution!

jeqcho:

Rishet11:

jeqcho (approved):
Thanks @Rishet11 — this is a well-scoped fix with an exemplary issue report to back it up (#255).

What I checked:

- **Bug is real.** On `main`, `src/inspect_robots/registry.py` (`_ensure_loaded`, lines ~101–105) sets `_loaded_builtins = True` and `_loaded_entrypoints = True` *before* the import/scan they guard, so a raising `import inspect_robots._builtins` or `entry_points()` scan permanently latches the "already loaded" path and every later call serves an empty registry. The repro in #255 confirms it.
- **Fix is correct and minimal.** Moving each flag assignment after its work turns "silently empty forever" into "retry / re-raise", which is the right behavior of the three. The per-plugin `try/except` in the entry-point loop is untouched, so a single broken plugin still warns by name rather than aborting discovery. A retried scan is also safe: `setdefault` keeps partial progress idempotent, and Python drops a failed module from `sys.modules`, so the builtins import retries cleanly.
- **Reentrancy considered.** The one behavioral trade-off of flag-after-work is that a reentrant `_ensure_loaded` call during discovery would re-enter the load path — but `sys.modules` guards the builtins import and `setdefault` guards the scan, so there's no recursion or duplication hazard in practice. Not a blocker.
- **Tests are solid.** Both new tests in `tests/test_registry_cli.py` are hygienic (all mutations go through `monkeypatch`, so global flags and `builtins.__import__` are restored), `sys` was already imported in the file, the builtins-import test correctly pins `_loaded_entrypoints = True` to isolate the branch under test, and the scan test verifies the retry actually happens (`len(calls) == 2`). Author confirmed both fail against unfixed code.
- **CI is fully green** across the matrix (lint, type, all OS/Python/NumPy combos, plugin jobs), and the CHANGELOG entry follows the existing format.

One take-it-or-leave-it nit: the CHANGELOG hunk introduces a double blank line before the new entry (`CHANGELOG.md`, around line 166) — feel free to fix or ignore.

Nice work — happy to merge.

jeqcho (approved):

PR #257, opened 2026-08-01
fix(view): render a non-finite metric instead of crashing on the null

### LINKED ISSUE #253: [bug] non-finite metrics are written as null and then crash inspect and view
**Describe the bug**

A non-finite metric (e.g. `min_distance_to_goal` = `inf`) is written to the log as JSON `null` **by
design** — `_sanitize` does this deliberately so the file stays RFC 8259 valid. But nothing on the
read side tolerates `None`, so every reader crashes on a log the sink itself produced.

`inspect-robots run` prints the metric fine, because it formats the in-memory log. The failure is
purely in the round-trip.

**To reproduce**

An adapter that simply doesn't publish `distance` in `StepResult.info` is enough — the scorer then
reduces to `inf`:

```python
import dataclasses, glob, json
from inspect_robots import eval as run_eval
from inspect_robots.registry import resolve
from inspect_robots.scorer import min_distance_to_goal

emb = resolve("embodiment", "cubepick")

class NoDistance:  # a conformant adapter that publishes no "distance"
    def __init__(self, i): self._i = i
    def __getattr__(self, n): return getattr(self._i, n)
    def reset(self, *a, **k): return self._i.reset(*a, **k)
    def step(self, *a, **k):
        r = self._i.step(*a, **k)
        return dataclasses.replace(r, info={k: v for k, v in r.info.items() if k != "distance"})

task = dataclasses.replace(resolve("task", "cubepick-reach"), scorer=min_distance_to_goal())
logs = run_eval(task, resolve("policy", "scripted"), NoDistance(emb), log_dir="/tmp/infprobe")

print("in-memory:", logs[0].results.metrics)                                   # {'min_distance_to_goal': inf}
print("on-disk  :", json.load(open(glob.glob('/tmp/infprobe/*.json')[0]))["results"]["metrics"])
                                                                               # {'min_distance_to_goal': None}
```

Then any reader:

```
$ inspect-robots inspect /tmp/infprobe/<log>.json
  File "src/inspect_robots/cli.py", line 1518, in _cmd_inspect
    print(f"  {name}: {value:.4g}")
TypeError: unsupported format string passed to NoneType.__format__

$ inspect-robots view /tmp/infprobe/<log>.json -o out.html
  File "src/inspect_robots/_html.py", line 270, in _number
    return f"{value:.4g}"
TypeError: unsupported format string passed to NoneType.__format__

$ inspect-robots view /tmp/infprobe -o idx
warning: could not read or render <log>.json: unsupported format string passed to NoneType.__format__
index: idx/index.html (1 logs, 0 pages, 0.0 MB)
```

**Expected behavior**

A saved log the framework wrote should always be readable by `inspect` and `view`. A non-finite
metric should render as something (`n/a`, or `inf` if the distinction is preserved), not crash.

The directory-index message is separately misleading: the log *is* readable — `read_eval_log`
succeeds — it just can't be formatted. It gets reported as unreadable and silently dropped from the
index.

**Additional context**

- Write side, intentional: `logging/json_log.py:41-52` (`_sanitize`), whose own docstring names
  `min_distance_to_goal` as the motivating case.
- Crash sites, all formatting with `:.4g` and none `None`-tolerant: `cli.py:1518`, `cli.py:1521`,
  `_html.py:270` (`_number`, typed `value: int | float`), `_html_index.py:148`.

Two possible fixes:
1. **Render side** (smaller, no schema change): make the four sites `None`-tolerant via one helper.
2. **Read side**: map `None` back to `float("inf")` for numeric score fields. This also fixes the
   silent information loss for API consumers of `read_eval_log` (`inf` in, `None` out) — but it's a
   schema decision, and `null` can't distinguish `inf` from `-inf`/`nan`, so I'd default to (1)
   unless you want the round-trip to be lossless.

Happy to send a PR for (1) if that's the preferred direction.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

Rishet11:
Fixes #253.

`_sanitize` (`json_log.py:41-52`) deliberately writes non-finite floats as JSON
`null` so the log stays RFC 8259 valid. That is correct and is unchanged here.
The problem is that nothing on the read side tolerated the `None` coming back, so
a log the sink itself wrote crashed every reader:

```
inspect-robots inspect <log>   TypeError: unsupported format string passed to NoneType.__format__
inspect-robots view <log>      same, at _html.py:270
inspect-robots view <logdir>   warning: could not read or render <log> ... (1 logs, 0 pages)
```

That last one was also misleading: the log reads fine, it just could not be
formatted, so it was reported as unreadable and silently dropped from the index.

After:

```
metrics:
  min_distance_to_goal: n/a
index: idx/index.html (1 logs, 1 pages, 0.0 MB)
```

**Scope.** Render side only — the four sites that format with `:.4g`
(`cli.py` metrics loop and per-scene `reduced` loop, plus the `_number` helper in
each of `_html.py` and `_html_index.py`), which render `None` as `n/a`. Eight
source lines.

I deliberately did **not** map `null` back to `inf` on read. That would make the
round-trip lossless for API consumers, but it is a schema decision and `null`
cannot distinguish `inf` from `-inf` from `nan`. Happy to go that way instead if
you prefer — it would close all four crash sites at once.

I also left the two near-identical `_number` helpers un-deduplicated, since
merging them touches files outside this bug.

**Tests.** Regression tests in `tests/test_registry_cli.py`, `tests/test_html_view.py`
and `tests/test_html_index.py`, covering `inspect`, single-file `view`, and the
directory index. All confirmed failing against the unfixed code.

Local: `pytest` 995 passed / 3 skipped, coverage 100%, `ruff check` clean,


jeqcho (approved):
Thanks for this, @Rishet11 — both the issue and the PR are a pleasure to review. The write-up traced the bug precisely (write side intentional, read side intolerant), the fix is exactly the minimal render-side change, and the regression tests cover all three affected surfaces (`inspect`, single-file `view`, and the directory index — including the misleading "could not read or render" drop, which was a nice catch on its own).

I verified the scope claim independently: the two `.4g` sites you left alone (`_print_run_summary` and `_print_eval_set_summary`) only ever format in-memory logs, where non-finite values are still real floats and format fine — and `eval_set` doesn't yet read logs back from disk. So the four sites you touched are indeed the complete set. Rendering `n/a` rather than guessing `inf` is also the honest choice, since `null` genuinely can't distinguish `inf`/`-inf`/`nan` on disk.

One tiny optional nit, no need to block on it:

1. The docstring in `tests/test_html_index.py`'s new test refers to `sanitize()` — the helper is `_sanitize` (your other test docstrings get it right). Fine to fix here or leave for a follow-up.

The two things you deliberately deferred (mapping `null` back on read for a lossless round-trip, and deduplicating the `_number` helpers) are reasonable follow-ups and I agree they don't belong in this PR. Merging as-is. Thanks again for the careful work!

Rishet11:
Thanks for the merge. Noting for anyone who finds this later: the lossless alternative (mapping `null` back to `inf` on read) is still on the table if the round-trip ever needs to be exact for API consumers — it just can't distinguish `inf` from `-inf` from `nan`, which is why this PR fixed the render side instead.

Issue #73, opened 2026-07-14
CLI reports status: success and exits 0 when every sample errored


aris-zhu:
## Symptom

Running the zero-config CLI with a broken setup (e.g. an invalid API key for the agent policy):

```
$ inspect-robots "place the fork on the plate" --policy agent -P model=anthropic/claude-fable-5 --embodiment cubepick
...
status: success
scenes: 1  trials: 1
log: logs/adhoc_....json
```

Exit code is 0 and the summary says success, yet every sample in the log errored:

```
samples[0].error: PolicyError: LLM request rejected — HTTP 401 ... "Invalid Anthropic API Key"
samples[0].score: None
```

The failure is only visible by opening the log file. Found while verifying #72 (a stub key produced a green-looking run).

## Why it happens

`eval()` follows Inspect-style `fail_on_error` semantics for `PolicyError` (`src/inspect_robots/eval.py:113`, default `False`): trial errors set the scene status and are excluded from scoring, but the top-level log `status` stays `"success"` unless the `fail_on_error` threshold trips (`eval.py:348-354`). Only `EmbodimentFault`/`SafetyAbort` (and reducer failures) force `status = "error"`.

The library default is defensible (long hardware runs should survive flaky trials), but the CLI neither exposes `fail_on_error` nor mentions errored samples in its run summary, so a 100%-errored run is indistinguishable at a glance from a healthy one.

## Possible directions (not mutually exclusive)

1. Print an errored-sample count in the CLI run summary (e.g. `trials: 1 (1 errored)`) whenever it is nonzero, and surface the first error message.
2. Expose `--fail-on-error[=THRESHOLD]` on the CLI, mapping to the existing `eval()` parameter.
3. Treat the degenerate all-samples-errored case as `status: error` / exit 1, at least in the zero-config path where the user just ran one instruction and it never produced a scored trial.

## Repro

With the fix from #72: put `ANTHROPIC_API_KEY=sk-invalid` in `./.env` (or export it), then run the command above and check `echo $?` and the printed status.

PR #258, opened 2026-08-01
fix(eval): measure the fail_on_error proportion against planned trials

### LINKED ISSUE #254: [bug] fail_on_error as a proportion always halts on the first error
**Describe the bug**

`fail_on_error` as a *proportion* (`0 < x < 1`) always halts on the very first error, making it
behaviourally identical to `fail_on_error=True`. The proportion is computed against the trials
completed **so far**, not the planned total, so the first errored trial is always `1/1 = 100%`,
which exceeds any threshold below 1.

This contradicts both the docstring and `run --help` ("`0<X<1` = proportion").

**To reproduce**

Unit level:

```python
from inspect_robots.eval import _should_fail

# "tolerate up to 50% errors"
print(_should_fail(0.5, 1, 1))   # True   <- first error, 1 trial so far => halts
print(_should_fail(0.5, 1, 4))   # False  <- what it should be for 1 error of 4 planned

for thr in (0.1, 0.5, 0.9, 0.99):
    print(thr, _should_fail(thr, 1, 1))   # all True
```

End to end: a 4-scene `cubepick-reach` with a policy that raises only on `scene-0`, run with
`fail_on_error=0.5` (tolerate up to 2 of 4 failures):

```
status error   error "fail_on_error threshold exceeded (1 errors)"
scenes recorded: [('scene-0', 'error')]   total_trials 1
```

The eval halts after the first trial and the remaining three scenes are never attempted, even
though 1/4 = 25% is under the 50% threshold.

**Expected behavior**

The run continues until the *proportion of planned trials* that errored reaches the threshold. With
`fail_on_error=0.5` over 4 scenes, it should take 2 errors to halt.

**Additional context**

`eval.py:452` passes `total_trials` — the running completed count — as the denominator into
`_should_fail` (`eval.py:537-546`, `errors / trials >= fail_on_error`).

The planned count is already derivable in scope: `len(task.scenes)` and the resolved epoch count.
Passing that as the denominator (or as a separate `planned` argument) fixes it without touching the
`True` / count (`x > 1`) branches, which behave correctly today.

This is the quietly-wrong kind of bug: the run produces a plausible-looking error log for an eval
that should have completed, so a user setting a tolerance gets a stricter policy than they asked
for and no indication of it.

Separately, and much smaller: the flag has no lower bound, so `--fail-on-error -1` also silently
behaves like `1` (falls through to `errors >= -1`). Happy to fold a `>= 0` check into the same PR
or keep it separate, whichever you prefer.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

Rishet11:
Fixes #254. Flagging up front that this **changes runtime behaviour**, so it may
want your judgement rather than a straight merge.

`_should_fail` received `total_trials` — the *completed so far* count — as the
denominator (`eval.py:452`). So the first errored trial was always `1/1 = 100%`,
which exceeds any threshold below 1, and `0<x<1` was behaviourally identical to
`fail_on_error=True`. Both the docstring and `run --help` document it as a
proportion.

```python
_should_fail(0.5, 1, 1)   # True   <- first error, 1 trial so far => halts
_should_fail(0.5, 1, 4)   # False  <- 1 error of 4 planned
```

End to end, a 4-scene task failing only scene-0 under `fail_on_error=0.5` halted
after one trial and never attempted the other three.

**Change.** The denominator is now `len(task.scenes) * epoch_spec.count`, both
already in scope. The `True` and count (`x > 1`) branches are untouched.

**On the existing test.** `test_fail_on_error_proportion_halts` still passes: it
uses a policy where every trial raises and asserts only `status == "error"`, which
holds either way — it now trips at 2 of 4 planned rather than 1 of 1. I updated
its comment to match and left the assertion alone. If that test was meant to pin
the old denominator, say so and I will close this.

**Not included:** the flag also has no lower bound, so `--fail-on-error -1`
silently behaves like `1`. Kept out to keep this PR to one concern; happy to
follow up.

**Tests.** `test_fail_on_error_proportion_tolerates_errors_below_the_threshold`
in `tests/test_rollout_hardening.py` — one error in four planned trials, asserts
all four scenes run. Confirmed failing against the unfixed code.

Local: `pytest` 991 passed / 3 skipped, coverage 100%, `ruff check` clean,


Rishet11:
Housekeeping: I opened five small fixes at once, and all five originally added their CHANGELOG entry at the same spot, so every pair collided. I've re-anchored them to different points in the `### Fixed` list — 8 of the 10 pairs now merge cleanly in any order.

The one remaining overlap for this PR is with #256. Whichever you merge second will want a one-line CHANGELOG rebase; ping me and I'll push it immediately. No code overlap, only the changelog line.

jeqcho (approved):
Thanks for this one — and for the unusually careful writeup on both the issue and the PR. The unit-level repro plus the end-to-end trace made it immediately verifiable, and flagging up front that this changes runtime behaviour (and why the existing test still passes) is exactly the context a reviewer needs.

The fix is right: the docstring, `run --help`, and `errors.py` all promise Inspect-style semantics, and Inspect measures the proportion against the task's planned total — so `len(task.scenes) * epoch_spec.count` is the correct denominator, and it's exact given the loop structure. I confirmed the `True` / count branches, the `EmbodimentFault`/`SafetyAbort` halts, and the all-errored safety net (#73) are all unaffected, and the new test genuinely discriminates against the old behaviour.

Two tiny, non-blocking notes:

1. `_BoomOncePolicy.act` builds a fresh `ScriptedPolicy()` on every call — works fine since its `act` is stateless given the observation, but holding one instance in `__init__` would be a touch cleaner. Happy to take it as-is.
2. The `--fail-on-error -1` lower-bound gap you mentioned is worth a follow-up PR — a `>= 0` validation with a clear error would round this out nicely. No need to fold it in here; keeping this PR to one concern was the right call.

Merging as-is. Thanks again — this is a model bug report + fix.

Rishet11:
Thanks for the merge and the detailed review.

Both notes picked up: the `--fail-on-error` lower bound is now #286. Scope grew slightly on evidence — `nan` is broken in the opposite direction (it fails every comparison, so the run never halts), so the guard is finite-and-`>= 0` rather than just `>= 0`. It lives in `_check_shared_run_conflicts` so `eval-set` gets it too.

On `_BoomOncePolicy.act`: agreed, holding one `ScriptedPolicy` in `__init__` is cleaner. I left it out of #286 to keep that PR to one concern — say the word and I'll send it as a one-liner.

PR #268, opened 2026-08-02
fix(setup): camera wizard UX follow-ups from #262 review (#263)

### LINKED ISSUE #263: setup: wizard camera UX follow-ups from the #262 review
Three non-blocking improvements deferred from the #262 fresh-eye review (all pre-existing behaviors that the device-identity inventory makes more visible):

1. Duplicate-camera detection compares strings, not physical identity. The already-assigned check in `_prompt_device_slot` compares selected device strings, so one physical camera can be assigned to two roles under different name families (by-id row for one role, a post-replug re-derived name for another). With the inventory now carrying `_CameraNode.camera`, the check could map both sides to the physical camera before comparing.

2. Listing rows are not refreshed after a `u` replug. Section row lists are computed once; after an unplug-identify replugs a camera, udev may have re-rolled its links, leaving a stale row that later slots still offer by number. Cheap fix: rebuild rows from a fresh inventory after each successful identify.

3. All-fallback-state cosmetics. When no camera has a usable by-id name, the slot prompt still advertises `p` although both views render identical rows, and the listing header says `under /dev/v4l/by-id` while showing by-path names. Suppress the toggle when the views are identical, or neutralize the header.

Refs #261, #262.

Adityakk9031:
Closes #263.

Addresses three non-blocking camera UX follow-ups deferred from the #262 review:

1. **Duplicate camera detection**: Maps candidate and assigned paths to physical camera identity (`_CameraNode.camera` / sysfs USB device directory or canonical resolved node) before comparing, catching duplicates across `by-id` and `by-path` aliases.
2. **Fresh listing row rescan after replug**: Rescans inventory and rebuilds `by_id_devices` and `by_path_devices` after a successful `u` unplug-to-identify flow, clearing stale rows.
3. **Toggle advertisement cleanup**: Suppresses the `p` prompt when `by_id_rows == by_path_rows`.

### Verification
- `ruff check .` — All checks passed.
- `ruff format --check .` — All 123 files formatted.
- `pytest tests/test_setup.py` — 113 passed, 43 skipped (gated on POSIX/symlinks).


Adityakk9031:

jeqcho (changes requested):
Thanks @Adityakk9031 — this lands all three follow-ups from #263 cleanly, and the physical-identity duplicate check in particular is exactly the shape the issue hoped for: mapping both sides through `_CameraNode.camera` with a resolved-path fallback is robust, and the by-id-vs-by-path duplicate test genuinely exercises it (it fails on `main`, as a good regression test should). The in-place `clear()/extend()` rebuild after `u` is also the right call, since the row lists are shared by reference across slots.

Two things I'd like addressed before merge, plus a couple of optional notes:

1. **`test_camera_view_state_suppresses_toggle_when_rows_are_identical` doesn't exercise the new code** (`tests/test_setup.py`, ~line 3819). Its inventory record has `by_id` set and a unique serial, so the pre-existing inventory branch already returns `advertise_path_toggle=False` — the test passes without this PR's change. To cover the new suppression line, use a record with `by_id=None` (and a by-path name): the old code advertises the toggle for it, both views then render the identical by-path row, and the new `by_id_rows == by_path_rows` guard is what flips it to `False`.

2. **Reprint the active listing when the post-`u` rescan changes the rows** (`src/inspect_robots/_setup.py`, ~lines 567-575). The numbered listing is printed once at section start; after a replug re-rolls udev links, the rebuilt rows can diverge from what's on screen, so a number typed at a later slot silently selects a different device than the one displayed — the very race this change is meant to close. A cheap guard: capture the old active rows, and if they differ after the rebuild, call `_print_camera_listing` for the active view.

Optional, non-blocking:

- A small test for the rescan path (scripted `u` with a `rescan_inventory` that returns changed rows, asserting the row lists were rebuilt) would protect item 2 of the issue, which currently has no coverage.
- `advertise_path_toggle` isn't recomputed after the rescan, so a replug that changes by-id availability leaves the `p` advertisement stale for later slots — fine to defer, just noting it.
- Tiny style nit: `physical_id` is re-defined on every loop iteration and the `kind == "v4l2" and inventory` guard appears twice; hoisting the helper above the loop would tidy this.

Also a heads-up that the branch conflicts with `main` (the `_duplicated_serials` → `_ambiguous_identities` refactor touched `_camera_view_state`), so the rebase will need a little care in exactly the function this PR edits.

Thanks again — with the test fixed and the listing reprint added, this is good to go.

Adityakk9031:
Hi @jeqcho, thanks for the thorough review! I have rebased onto `main` and addressed all feedback:

1. **Camera view state suppression test**: Updated `test_camera_view_state_suppresses_toggle_when_rows_are_identical` to use `by_id=None` with by-path rows, which would otherwise advertise the toggle due to missing by-id names and thus directly exercises the new `by_id_rows == by_path_rows` suppression guard.
2. **Reprint active listing on post-`u` rescan changes**: Captured `old_active_rows` and call `_print_camera_listing(new_active_rows, device_dir, out)` if the active rows changed after rescan.
3. **Rescan coverage**: Added `test_prompt_device_slot_u_rescan_rebuilds_and_reprints_on_row_change` verifying that unchanged rescans do not reprint, while changed rescans rebuild rows and reprint the active listing.
4. **Cleanups**: Hoisted `physical_id` helper above the loop in `_prompt_device_slot`, simplifying duplicate checking across kind and inventory states.
5. **Rebased**: Rebased onto `main` cleanly integrating the ambiguous-identities refactor.

All checks (ruff lint/format, Linux-targeted mypy, and unit tests) pass.

Adityakk9031:

1. **Toggle suppression logic:** Refined the toggle suppression condition in `_camera_view_state` so the `'p'` port-view hint is only suppressed when already in port view (`not active_is_by_id`) and by-id/by-path rows match, preventing contradictory prompts.
2. **Tests & CI:** All tests in `tests/test_setup.py` and all CI checks are passing (`ci-ok pass`).

PR #269, opened 2026-08-02
fix(rerun): shut down live worker on eval start to prevent blueprint …

### LINKED ISSUE #267: rerun sink: blueprint state races on the crash-path sink-reuse pattern
Deferred from the #266 review (both pre-existing-class lifecycle races, no crash, narrow window):

1. If an eval ends without on_eval_end (Ctrl-C paths the sink docstring already documents) and the caller reuses the sink, the still-draining previous worker can interleave with the next eval's bind_spaces writes and on_eval_start blueprint reset: a leftover payload may build a blueprint from a torn mix of old/new space fields, or the worker's _blueprint_prefix write can land after the reset so the next eval's first trial silently gets no layout. Candidate fix: shut down a live worker at on_eval_start before resetting (also bounds the pre-existing step-data contamination on that path), or guard the blueprint fields with the sink condition lock.

2. A worker disowned while blocked in send_blueprint can set _blueprint_warned after the next eval's reset, suppressing that eval's single fallback warning (same cosmetic class as the existing _emit_warned race).

Refs #265, #266.

Adityakk9031:
Closes #267.

### Problem
When an eval ends without `on_eval_end` (e.g. Ctrl-C or crash path) and the caller reuses the `RerunSink` instance, the still-running worker thread from the previous eval can interleave with `bind_spaces` and `on_eval_start` blueprint resets on the subsequent eval. This could cause torn blueprint payloads or suppressed blueprint warnings.

### Fix
In `on_eval_start`, check if a worker thread is still alive and shut it down before resetting `_blueprint_prefix` and `_blueprint_warned`.

> **Note on residual race**: The disowned-worker case (a worker wedged in `send_blueprint` past `flush_timeout`) can still set `_blueprint_warned` after the reset, so cosmetic race 2 from #267 is narrowed rather than fully eliminated.

### Verification
- `ruff check .` — Passed.
- `ruff format --check .` — Passed.
- `pytest tests/test_rerun_sink.py` — 66 passed, 4 skipped, 0 failures.


Adityakk9031:

jeqcho (changes requested):
Thanks for picking up #267 so quickly — this lands exactly the candidate fix the issue sketched, and it's in the right place: stopping a leftover worker before the blueprint reset (and before `rr.init` re-points the global recording) closes the reset interleave and bounds cross-eval step contamination. The `is_alive()` guard is a nice touch too, since it keeps the normal start path from paying the `_probe_recording_flush` timeout. I also verified the new test fails without the source change and passes with it, so it genuinely guards the fix.

A few small asks before merge, all in the test:

1. The hand-rolled `FiringRR` fake is missing `Scalars`/`Scalar`, so the worker's emit actually raises `AttributeError` inside the test and triggers the "RerunSink failed to emit a step" warning (you can see it in the pytest output), and `FiringRR.events` never receives anything. Please use the file's existing `_install_fake_rerun(monkeypatch)` + `_log_one(sink, t)` helpers instead — they simulate a working SDK surface, avoid the stray warning, and will roughly halve the test.
2. Placement: the test currently sits between `test_real_rerun_accepts_the_blueprint` and `test_real_rerun_process_exits_when_tcp_peer_never_reads`, which are the real-SDK group at the end of the file. It fits better next to `test_eval_end_shuts_down_worker_and_log_step_restarts_it` with the other lifecycle tests.
3. This module comments its non-obvious moves pretty thoroughly — a one-liner above the new guard (e.g. "a crash path can skip `on_eval_end`; stop the previous eval's worker before resetting blueprint state so its writes can't land after the reset") would help future readers.
4. Optional, but worth a sentence in the PR description: the disowned-worker case (a worker wedged in `send_blueprint` past `flush_timeout`) can still set `_blueprint_warned` after the reset, so #267's cosmetic race 2 is narrowed rather than fully eliminated. Fine to still close the issue given how it was framed — just good to state the residual explicitly.

The core change is solid; happy to approve once the test is reworked.

Adityakk9031:
Thanks for the detailed review @jeqcho!

I have addressed all the feedback and rebased the branch onto the latest `main`:

- **Test helpers:** Refactored the test to use `_install_fake_rerun(monkeypatch)` and `_log_one(sink, 0)`, eliminating the stray warning.
- **Test placement:** Moved `test_sink_reuse_after_interrupted_eval_shuts_down_previous_worker` to sit alongside the lifecycle test group next to `test_eval_end_shuts_down_worker_and_log_step_restarts_it`.
- **In-code documentation:** Added the explanatory comment above the `_shutdown()` check in `on_eval_start`.
- **Merge conflict & PR description:** Resolved the conflict with upstream in `on_eval_start` and updated the PR description to explicitly document the disowned-worker residual note.

jeqcho (approved):
Thanks @Adityakk9031, this round lands everything from the last review cleanly. The test is now the compact lifecycle test I was hoping for: it reuses `_install_fake_rerun` and `_log_one`, so the worker actually emits through a working fake SDK surface and the stray "failed to emit a step" warning is gone (I ran it 30 times under `-W error` with zero failures). It sits directly after `test_eval_end_shuts_down_worker_and_log_step_restarts_it`, where the other lifecycle tests live, and the comment above the guard says exactly why the shutdown happens before the reset. The rebase also came through without collateral: the source diff against main is still just the four lines in `on_eval_start`, and `origin/main` has not touched `rerun_sink.py` since your merge base, so nothing was dropped or reintroduced.

Prior review points:

1. ✅ Test uses the file's helpers instead of a hand-rolled fake. `tests/test_rerun_sink.py:1549-1551`: `_install_fake_rerun(monkeypatch)` / `RerunSink()` / `_log_one(sink, 0)`. No warning in the pytest output any more.
2. ✅ Placement. `tests/test_rerun_sink.py:1545` is immediately after `test_eval_end_shuts_down_worker_and_log_step_restarts_it` (ends at `:1542`) and before `test_emit_failure_warns_once_and_keeps_running`.
3. ✅ In-code comment. `src/inspect_robots/logging/rerun_sink.py:651-652`: "A crash path can skip on_eval_end; stop the previous eval's worker before resetting blueprint state so its writes can't land after the reset."
4. ✅ PR description states the residual. The "Note on residual race" paragraph covers the disowned-worker case for `_blueprint_warned`.

What I verified on this head:

- Reverted only the `rerun_sink.py` hunk and ran the new test: it fails at `tests/test_rerun_sink.py:1558` with `assert not prev_worker.is_alive()`, so it genuinely guards the fix.
- Full gates: `ruff check`, `ruff format --check`, `mypy` (strict, src + tests) all clean; `pytest --cov` at 100.00% line and branch with the CI `--extra dev` environment; the file also passes against the real rerun-sdk 0.34.0 as `test-rerun` will run it.
- Behaviour: the `is_alive()` guard means the normal path (previous eval ended via `on_eval_end`, `_worker` is `None`) pays nothing. On the crash-reuse path `_shutdown()` runs before `rr.init` re-points the global recording, so the `_probe_recording_flush` probe hits the previous recording, which is the one whose flush we care about. A stalled probe disables the sink exactly as it would have at `on_eval_end`, so no new failure mode. No hardware-side effect: this is a visualization sink and the added wait is bounded by `flush_timeout` before the rollout starts.
- Merge state: branch is behind main by 13 commits but `merge-tree` is clean, so the "update branch" step before merge will be a no-op conflict-wise.

Non-blocking notes, no action needed for this PR:

- `eval()` calls `bus.bind_spaces(...)` at `src/inspect_robots/eval.py:364` before `bus.on_eval_start(spec)` at `:366`, so a worker still draining leftover payloads can technically overlap the new eval's `bind_spaces` writes before your guard fires. Any torn blueprint it builds goes out under the old prefix to the old recording, and the new eval's first payload re-sends from fully bound fields after the reset, so this is the same cosmetic class as the residual you already documented. Worth knowing, not worth widening this PR.

Approving. Nice work following through on the details.

PR #271, opened 2026-08-02
feat(agent): tag the final action of each chunk with meta['chunk_fina…

### LINKED ISSUE #167: agent policy: tag the final interpolant of a chunk so embodiments can settle only where it matters
## Context

`inspect-robots-yam#62` covers embodiments observing before commanded motion completes. The fix landing there settles on **every** control step, because the embodiment has no way to know which action is the last of a chunk.

Settling every step may be wasteful. The `agent` policy interpolates a tool call at 1% of joint range per step (`plugins/inspect-robots-agent/src/inspect_robots_agent/_tools.py:468`), and 1% of a ±π range is about 0.063 rad. That is a genuine motion per step, so converging on each one could turn a smooth 10 second sweep into 100 stop-and-go moves, which would also make the `_success` note's advertised playout duration (`_tools.py:356-364`) misleading.

Only the final interpolant's observation ever reaches the LLM, since `DefaultController` calls `policy.act()` only when its buffer empties (`src/inspect_robots/controller.py:57-62`). So that is the only step whose settling actually matters.

## Proposal

Have the agent policy tag the last `Action` of each chunk via `Action.meta`, e.g. `meta["chunk_final"] = True`. Embodiments that implement settling can then do it only on tagged actions and skip it elsewhere.

This needs **no core contract change**: `meta` is set policy-side, and `rollout` already reads a policy-set meta key (`meta["request_stop"]`, see `src/inspect_robots/rollout.py`). Documenting `chunk_final` as a recognized convention alongside it would be the only framework-level work, and even that is optional to start.

## Caveat

Whether this is worth doing depends on a measurement nobody has taken: how fast the servo closes 0.063 rad relative to the 100 ms control period. If it converges well inside the period, per-step settling is free and this is unnecessary complexity. The yam PR emits per-step settle residuals and timeout counts in `StepResult.info` precisely so that measurement becomes possible on hardware.

Filing now so the option is recorded, not to claim it should be built yet.

Adityakk9031:
Closes #167.

### Summary
Tags the final `Action` object of each motion chunk generated by agent policy tools (`move_to`, `move_by`, `move_joints`) with `action.meta["chunk_final"] = True`.

Embodiments and controllers can use this metadata to perform hardware sensor/joint settling exclusively on the final step whose observation reaches the LLM policy, avoiding wasteful stop-and-go delays on intermediate interpolation waypoints.

### Verification
- `ruff check .` — Passed.
- `ruff format --check .` — Passed.
- `pytest plugins/inspect-robots-agent/tests/test_tools_motion.py` — 92 passed.


Adityakk9031:

jeqcho (approved):
Thanks for picking this one up, @Adityakk9031 — this is a tidy, faithful implementation of #167.

What I checked:

- **Single tagging point covers all three tools.** Both chunk-producing paths (`_move` at `_tools.py:474` and the displacement path at `_tools.py:510`) funnel through `_success`, so tagging there correctly covers `move_to`, `move_by`, and `move_joints`.
- **Frozen-dataclass handling is right.** `Action` is frozen, and rebuilding the last element with `{**last.meta, "chunk_final": True}` preserves any existing meta rather than clobbering it.
- **Test asserts both sides** — intermediates untagged, final tagged — using the existing `_execute_absolute` helper, consistent with the file's conventions.
- **CI is fully green**, including the `plugin · agent` job.

Two non-blocking observations, fine to leave for follow-ups:

1. The `_stop` path (`_tools.py:326-342`) emits a one-action chunk that stays untagged. Since that action requests a stop and the trial ends, settling there is moot, so I think skipping it is defensible — just noting it in case anyone later relies on "every chunk's last action is tagged" as an invariant.
2. #167 mentions optionally documenting `chunk_final` as a recognized convention alongside `request_stop`. Not done here, and the issue itself calls it optional, so no action needed — the maintainers may want to add a line when an embodiment first consumes the tag.

Nice, minimal change. Thanks again!

jeqcho (approved):

PR #279, opened 2026-08-04
Operator console: live feedback to agent policies, explicit end/score commands

### LINKED ISSUE #273: Operator console: live feedback to agent policies, explicit end/score commands
## Problem

During an attended run the operator can see things the policy cannot ("the gripper is farther than it looks", "center the object first"), but there is no way to tell the policy anything while the trial runs. Worse, the current end-of-episode signal is any line on stdin (the yam embodiment's `default_poll_end` reads stdin and returns `terminated=True, termination_reason="operator_end"`), so touching the keyboard is how you end the run. Typing is unavailable as an input channel by construction.

## Proposal

Add an operator console to the framework, active only for attended TTY runs where the policy opts in.

**Feedback channel.** A non-empty line followed by Enter queues a message. The rollout injects queued messages into the policy-facing observation under a reserved `extra["operator_messages"]` key (list of `{"t", "text"}`), with the same delivered-once-per-inference discipline as `extra["approvals"]` (#217). The agent plugin renders them in the next model turn ("operator: ..."). Messages also land in the trial transcript so logs, the HTML view, and summarize can show what the operator said and when.

**End and score.** Since typed text now means feedback, ending needs an explicit but short gesture:

- bare Enter (empty line): end the episode with `operator_end`, verdict prompted afterwards exactly as today. This keeps the current muscle memory: pressing Enter still ends the run.
- `/y [note]`, `/n [note]`, `/p [note]`: end and record the verdict (and optional grader note) in one line, skipping the post-run prompt.
- any other `/...` line: print a one-line usage hint, do not queue it as feedback (protects against typos ending up in the model context).

The console owns the end signal, so the rollout, not the embodiment, terminates the trial with `OPERATOR_END` when the console is active.

**Gating.** Policies opt in via a duck-typed `accepts_operator_messages` attribute; the agent plugin sets it. VLAs never see the console, and their runs keep the current embodiment-owned keypress behavior. Unattended runs (`--no-prompt`, no TTY) are unchanged.

**stdin ownership.** Two readers on one stdin race, so when the console is active any embodiment that polls stdin for end-of-episode has to stand down (duck-typed hook; the yam plugin change is a follow-up issue in that repo). Mock and sim embodiments do not read stdin, so nothing changes for them.

## Scope

Core (`rollout`, `cli`, new console module, transcript), plus the agent plugin rendering. Design doc to follow in `plans/`.

jeqcho:
Closes #273.

Adds an operator console for attended runs where the policy opts in (the `agent` plugin does): a typed line + Enter is delivered to the policy at its next inference as feedback, bare Enter ends the episode with `operator_end` exactly as today's keypress does, and `/y` `/n` `/p [note]` end the episode and record the verdict in one stroke, skipping the post-trial prompt. VLA policies and unattended runs are unchanged.

Design doc: `plans/0042-operator-console.md` (two fresh-context critique rounds before implementation).

## How it works

- `console.py` (new): a threadless, non-blocking console. `poll()` does a zero-timeout `select()` and reads pending bytes at the fd level (`os.read`), assembling complete lines itself. No background thread means stdin keeps a single reader, so the existing post-trial verdict `input()` prompt still works. EOF latches permanently, including inside `begin_trial()`'s drain, so a redirected stdin can never spin.
- `rollout()` takes an optional `operator_input`. Messages are injected under the reserved `extra["operator_messages"]` key with the same delivered-once-per-inference tail discipline as `extra["approvals"]` (#217), and recorded as `operator_message` transcript events. An end request terminates the trial with `OPERATOR_END` before any inference or motion that step. A raising `poll()`/`begin_trial()` degrades with one warning and never costs the eval log.
- The log gains `SceneResult.operator_messages`, strictly parallel to `epochs` (two-level tuple coercion on read; additive, no schema bump). Summarize and the HTML view render it.
- CLI gating: attended TTY + policy `accepts_operator_messages` + POSIX + a console-safe embodiment, meaning it offers the duck-typed `defer_operator_end()` hook (called so stdin-polling embodiments stand down) or is simulated. Real hardware without the hook keeps today's behavior with a notice, so there is no two-readers-on-stdin race with unpatched embodiments.
- The `agent` plugin opts in, renders each message as `operator feedback (step N): ...` in the next observation, and both system templates tell the model to treat the lines as trusted guidance.

## Interim behavior on yam hardware

The console stays off (notice printed) until `inspect-robots-yam` implements `defer_operator_end()`; a follow-up issue will be filed there. Core can ship first with no release-ordering hazard.

## Testing

All gates green locally: ruff check, ruff format, mypy strict, `pytest --cov` at 100% core coverage (1075 passed), agent plugin suite (447 passed).

### DIFF
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -111,6 +111,23 @@ never adopts an embodiment verdict or prompts for any other ending, so
 unattended runs stay non-blocking. An unjudged trial honestly scores as
 failure with "no operator judgement recorded".
 
+### Live operator feedback
+
+On an attended run, an opted-in policy can also receive feedback while the
+episode is running. The CLI prints the operator-console usage hint when this
+channel is active. Type a normal line and press Enter to deliver it at the
+policy's next inference. Bare Enter ends the episode, while `/y`, `/n`, or `/p`
+plus an optional note ends it and records the verdict immediately, without a
+second post-trial prompt. Feedback is saved per trial and appears in summaries
+and HTML reports. Piped stdin and `--no-prompt` disable the channel.
+
+The console activates only where stdin has one owner: POSIX platforms and
+either a simulator or a real-hardware embodiment that offers
+`defer_operator_end()`. That hook tells an embodiment to stop polling its own
+end-of-episode keypress while the framework console owns stdin. Older hardware
+embodiments keep their existing keypress behavior, print a notice, and leave
+feedback typing off until they add the hook.
+
 ## `inspect-robots setup`
 
 The interactive first-run wizard: it prompts for each `[defaults]` key with
--- a/plans/0042-operator-console.md
+++ b/plans/0042-operator-console.md
@@ -1,0 +1,169 @@
+# Operator console — live feedback to agent policies Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** During an attended run with an opted-in policy (the `agent` plugin), the operator can type a line of feedback + Enter and have it delivered to the policy at its next inference ("the gripper is farther than it looks"), can end the episode with a bare Enter (unchanged muscle memory), and can end-and-score in one stroke with `/y`, `/n`, `/p` plus an optional grader note (#273). VLA runs and unattended runs are byte-for-byte unchanged.
+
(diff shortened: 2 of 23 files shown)


jeqcho:
Re the code-quality bot: leaving both as is. Protocol bodies use `...` everywhere else in this repo (sink.py, policy.py), and that is the standard typing idiom, so OperatorInput stays consistent with them. The mixed import style in test_registry_cli.py predates this PR; the new tests follow the file's existing monkeypatch pattern (`monkeypatch.setattr(inspect_robots, "eval", ...)`), and switching one function to a different style would make the file less uniform, not more.

PR #286, opened 2026-08-04
fix(cli): reject an out-of-range --fail-on-error

Rishet11:
The follow-up you suggested in #258.

`--fail-on-error` was `type=float` with no bounds, and out-of-range values were
not inert — they were silently reinterpreted:

- **negative** falls through to `errors >= fail_on_error`, so `--fail-on-error -1`
  halted on the first error exactly like `1`
- **NaN** fails every comparison, so the run never halted at all

```
--fail-on-error -1   ->  --fail-on-error must be finite and >= 0, got -1.0
--fail-on-error nan  ->  --fail-on-error must be finite and >= 0, got nan
--fail-on-error 0    ->  run status: completed        (still valid)
```

**Placement.** In `_check_shared_run_conflicts` rather than `_cmd_run`, since
unlike `--max-steps` this flag is shared by `run` and `eval-set` — so both get
the guard from one check. The message follows the neighbouring
`--max-action-delta must be finite and > 0, got ...` form, and `inf` is rejected
for the same reason that guard rejects it.

`0` remains valid; it is the documented "never halt" value, so it is deliberately
absent from the rejected list (and that is called out in the test docstring).

**Tests.** Parametrized over `-1`, `-0.5`, `inf`, `nan` for both `run` and
`eval-set`, mirroring `test_cli_run_rejects_invalid_max_action_delta` /
`test_cli_eval_set_rejects_invalid_max_action_delta` next door. All 8 confirmed
failing against the unfixed code.

Local: `pytest --cov` 1144 passed / 5 skipped at 100%, `ruff format --check` clean,
`mypy` clean. `ruff check` shows only the 2 pre-existing agent-plugin union-order
errors already on `main`.

I left the `_BoomOncePolicy.act` nit alone rather than fold a second concern in —
happy to send it as a one-liner if you want it.


jeqcho (approved):
Thanks for the careful follow-up from #258 — this one was a pleasure to review because the PR description made every claim easy to verify.

What I checked:

- **The bug is real.** In `src/inspect_robots/eval.py`, `_should_fail` sends a negative straight through `not fail_on_error` (falsy only for 0) and past the `0 < x < 1` band into `errors >= fail_on_error`, so `-1` genuinely halts on the first error like `1`. And NaN fails every comparison (`0 < nan < 1` and `errors >= nan` are both False), so the run silently never halts — good catch that this is broken in the *opposite* direction, which justifies the finite check rather than just a lower bound.
- **The fix is complete for the CLI surface.** `--fail-on-error` is defined in `_add_shared_eval_args`, and `_check_shared_run_conflicts` is called by both `run` and `eval-set` before defaults are loaded or components resolved, so both commands fail fast. The `is not None` gate is right: unset falls through to the existing `else False` at the call sites, and `0` (the documented "never halt" value) stays valid.
- **Idiomatic.** The guard sits directly beside the `--max-action-delta` check in `src/inspect_robots/cli.py` and mirrors its exact message shape and rationale comment; rejecting `inf` matches that neighbor's reasoning. `math` was already imported.
- **Tests are adequate.** The two parametrized tests in `tests/test_registry_cli.py` cover `-1`, `-0.5`, `inf`, `nan` on both commands, mirror the `max_action_delta` tests next door, and the docstring explicitly documents why `0` is absent from the list — nice touch.

One non-blocking observation: the programmatic `evaluate(fail_on_error=...)` API still accepts out-of-range values. That's consistent with how the neighboring `--max-action-delta` guard is scoped (CLI input validation), so I don't think it belongs in this PR — just noting it in case a library-level guard is ever wanted.

LGTM. Thank you!

jeqcho (approved):

PR #288, opened 2026-08-04
fix(approver): shape an explicit max_delta like the action box

### LINKED ISSUE #287: [bug] DeltaLimitApprover ignores box shape for an explicit max_delta, breaking multi-dimensional action spaces
**Describe the bug**

`_validate_max_delta` always returns a flat `(dim,)` array. The derived-default path
reshapes it to the box shape, with a comment saying exactly why
(`approver.py:230-232`):

```python
# review() clips against the action's own shape; a flat delta
# would fail to broadcast for multi-dimensional boxes.
self._delta = self._delta.reshape(action_space.shape)
```

The explicit-`max_delta` path never does. So for any action `Box` whose `shape` is
not 1-D (e.g. a bimanual `(2, 7)`), passing `max_delta` breaks the limiter in one
of two ways depending on control mode:

1. **Absolute modes** (`joint_pos`, `eef_abs_pose`): construction succeeds, so the
   guardrail reports itself as active, then `review()` raises a raw numpy
   broadcast `ValueError` mid-rollout on the *second* control step (the first has
   no reference yet, so it passes).
2. **Displacement modes** (`joint_delta`, `joint_vel`, `eef_delta_pos`,
   `eef_delta_pose`): the broadcast failure happens in `__init__` via `_intersect`,
   so `_build_guardrails` catches the `ValueError` and silently drops the limiter,
   surfacing an internal numpy message as the "reason". The requested
   `--max-action-delta` is not applied at all.

Case 2 is the one I'd call the real problem: `_build_guardrails`' own docstring
says the CLI is "never *less* protective than running without guardrails — and
never silently unprotected either", and here a user who explicitly asked for a
0.1 per-step limit gets a 1.0 box clamp and a warning that reads like a bug
report rather than a reason.

Multi-dimensional action boxes look intended, not out of contract: `Box.shape` is
`tuple[int, ...]`, `low`/`high` are validated against the full shape, `dim` is the
flattened product, the derived path deliberately handles N-D, and
`ActionSemantics.dim_labels` documents the bimanual `("left_j0", ...,
"right_gripper")` case. Every in-repo space happens to be 1-D, which is why this
is invisible today.

**To reproduce**

```python
import numpy as np

from inspect_robots.approver import DeltaLimitApprover
from inspect_robots.spaces import ActionSemantics, Box
from inspect_robots.types import Action

shape = (2, 7)  # bimanual: 2 arms x 7 joints
space = Box(
    shape=shape,
    low=np.full(shape, -1.0),
    high=np.full(shape, 1.0),
    semantics=ActionSemantics(control_mode="joint_pos"),
)

approver = DeltaLimitApprover(space, max_delta=0.1)
print("delta shape:", approver._delta.shape, "space shape:", space.shape)

store: dict = {}
approver.review(Action(data=np.zeros(shape)), store)  # step 1: sets reference
approver.review(Action(data=np.full(shape, 5.0)), store)  # step 2: must clamp
```

```
delta shape: (14,) space shape: (2, 7)
  File "/tmp/issue_repro.py", line 20, in <module>
    approver.review(Action(data=np.full(shape, 5.0)), store)  # step 2: must clamp
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../inspect_robots/approver.py", line 265, in review
    approved = np.clip(data, ref - self._delta, ref + self._delta)
                             ~~~~^~~~~~~~~~~~~
ValueError: operands could not be broadcast together with shapes (2,7) (14,)
```

And through the CLI's own guardrail builder, showing both modes:

```python
import numpy as np

from inspect_robots.cli import _build_guardrails
from inspect_robots.spaces import ActionSemantics, Box
from inspect_robots.types import Action

shape = (2, 7)
for mode in ("joint_pos", "joint_delta"):
    space = Box(
        shape=shape,
        low=np.full(shape, -1.0),
        high=np.full(shape, 1.0),
        semantics=ActionSemantics(control_mode=mode),
    )
    approver, active, warnings = _build_guardrails(space, 0.1)
    print(f"--- control_mode={mode!r}, --max-action-delta 0.1")
    print("  guardrails active:", active)
    print("  warnings:", warnings)
    store: dict = {}
    approver.review(Action(data=np.zeros(shape)), store)
    try:
        out = approver.review(Action(data=np.full(shape, 5.0)), store)
        print("  step 2 max |value|:", float(np.abs(np.asarray(out.data)).max()))
    except Exception as e:
        print(f"  step 2 CRASH: {type(e).__name__}: {e}")
    print()
```

```
--- control_mode='joint_pos', --max-action-delta 0.1
  guardrails active: ['clamp', 'delta-limit']
  warnings: []
  step 2 CRASH: ValueError: operands could not be broadcast together with shapes (2,7) (14,)

--- control_mode='joint_delta', --max-action-delta 0.1
  guardrails active: ['clamp']
  warnings: ['delta limit skipped: operands could not be broadcast together with shapes (2,7) (14,) ']
  step 2 max |value|: 1.0
```

**Expected behavior**

An explicit `max_delta` should behave like the derived default on the same box:
broadcast to `action_space.shape` and limit each dimension to `±0.1` from the last
approved action. For the `joint_pos` case above, step 2 should clamp to `0.1`
rather than raise. For `joint_delta`, `delta-limit` should be active with no
warning, and step 2 should report `0.1` rather than the box's `1.0`.

A genuinely inapplicable space should still be refused with the existing kind of
explanatory `ValueError`, not a numpy broadcast message.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

**Additional context**

Happy to send the fix: reshape the validated explicit delta to
`action_space.shape` once in `_validate_max_delta`'s caller so both the absolute
and displacement branches get it, plus regression tests covering a non-1-D box in
each mode. Existing behaviour for 1-D boxes is unchanged (`reshape((dim,))` is a
no-op there).

Rishet11:
Fixes #287.

## What was wrong

`_validate_max_delta` broadcast `max_delta` to a flat `(dim,)` array. The
derived-default path reshaped its delta to `action_space.shape` and said why:

```python
# review() clips against the action's own shape; a flat delta
# would fail to broadcast for multi-dimensional boxes.
self._delta = self._delta.reshape(action_space.shape)
```

The explicit path never did, so on any action `Box` whose `shape` is not 1-D
(a bimanual `(2, 7)`, say) the limiter broke in one of two ways:

- **Absolute modes** (`joint_pos`, `eef_abs_pose`): construction succeeded, so
  the guardrail banner reported `delta-limit` active, and then `review()` raised
  a raw numpy broadcast `ValueError` mid-rollout on the *second* control step
  (the first has no reference yet, so it passes).
- **Displacement modes** (`joint_delta`, `joint_vel`, `eef_delta_pos`,
  `eef_delta_pose`): the broadcast failed in `__init__` via `_intersect`, so
  `_build_guardrails` caught the `ValueError` and dropped the limiter, showing an
  internal numpy message as the "reason". `--max-action-delta 0.1` on a
  `±1.0` box then limited to `1.0`.

The second is the one I'd weight, because `_build_guardrails`' own docstring says
the CLI is "never *less* protective than running without guardrails — and never
silently unprotected either".

## The fix

`_validate_max_delta` now takes the box `shape` and returns the validated delta
reshaped to it, so both the absolute branch and the displacement `_intersect`
broadcast against the action. One shaping site instead of two, next to the
validation it belongs with. `reshape((dim,))` is a no-op for 1-D boxes, so
existing behaviour is unchanged — including the existing `ValueError` messages
for `max_delta=0.0`, `inf`, a negative entry, and a wrong-length array, which are
still raised before the reshape is reached.

## Tests

Two regression tests, one per branch, both asserting the clamped **values** and
`meta["delta_clamped"]` rather than just absence of an exception. Against
unfixed code they fail in the two different places the bug lives:

- absolute → `approver.py` `review()`, `np.clip(data, ref - self._delta, ...)`
- displacement → `approver.py` `_intersect`, at construction

## Verification

```
pytest --cov   1138 passed, 5 skipped   Total coverage: 100.00%
```

- **No new `--max-action-delta` CLI validation.** The flag's finite/`> 0` check
  already exists; this was purely a shape bug, so the flag surface is unchanged.
- **No change to `_build_guardrails`' catch-`ValueError`-and-warn behaviour.**
  Turning a broadcast bug into a clearer warning would have hidden the actual
  defect. That degradation path is correct in principle, and the honest fix is
  for the constructor not to fail on a valid space. Happy to revisit separately
  if you'd like it to distinguish "cannot apply to this space" from "internal
  error" — that felt like a design call rather than part of this fix.
- **No multi-dimensional built-in embodiment or fixture.** Every in-repo space is
  1-D, which is exactly why this was invisible; the tests construct the `(2, 7)`
  box directly rather than adding a bimanual mock.
- **Left `frames.py` alone.** Unrelated, just noting I didn't wander.

One note on reachability, so it's not buried: no built-in embodiment has a
non-1-D action space, so this needs a custom or adapter embodiment (a bimanual
arm) to hit in practice. I reproduced it through `_build_guardrails`, the CLI's
real code path, not through an end-to-end `inspect-robots run`. If you consider
N-D action boxes out of contract, then the interesting part of #287 is the
displacement case silently dropping a requested guardrail, and I'd be glad to
narrow the PR to that.


jeqcho (approved):
Thanks @Rishet11 — this is a model bug report and fix pair. The issue writeup made it easy to confirm, and the PR does exactly what it says with no scope creep.

What I checked:

- **The bug is real.** Reproduced both failure modes against `main`: on a `(2, 7)` box with `max_delta=0.1`, absolute mode crashes with a raw numpy `ValueError` in `review()` (`src/inspect_robots/approver.py:265` on main), and displacement mode fails in `_intersect` at construction, which makes `_build_guardrails` silently drop the limiter a user explicitly asked for. The displacement case is indeed the serious one given the guardrail builder's "never silently unprotected" contract.
- **The fix is right and minimal.** Reshaping once in `_validate_max_delta` (`src/inspect_robots/approver.py:293`) covers both branches, mirrors the derived-default path's existing reshape (with a matching comment explaining why), and is a no-op for every 1-D in-repo space. All four existing validation errors (`0.0`, `inf`, negative, wrong length) still fire before the reshape is reached, so messages are unchanged.
- **Tests fail for the right reasons.** I ran the two new regression tests against main's `approver.py`: both fail, each at the distinct site the PR description names, and both assert clamped values plus `meta["delta_clamped"]` rather than mere non-raising. Full `tests/test_approvers.py` passes on the branch (35/35), and CI is green across the matrix, including the `numpy>=1.24,<2` leg.
- **Changelog** entry sits correctly under `[Unreleased]` → `Fixed`.

Two non-blocking notes:

- `_validate_max_delta` now takes both `shape` and `dim` even though `dim == prod(shape)`; deriving `dim` inside would trim the signature. Keeping it as-is preserves the existing error text, so entirely fine either way.
- A per-dimension `max_delta` for an N-D box must still be passed flat `(dim,)` — a `(2, 7)`-shaped array is rejected by the broadcast-to-`(dim,)` contract. That's pre-existing behavior and consistent with `dim_labels` flattening, just worth remembering if N-D embodiments become first-class.

The "what I deliberately left out" section was appreciated — agreed that distinguishing "inapplicable space" from "internal error" in `_build_guardrails` is a separate design question. Merging as-is.

jeqcho (approved):

PR #290, opened 2026-08-04
fix(task): reject duplicate scene ids

### LINKED ISSUE #289: [bug] Duplicate scene ids silently overwrite each other's stored frames
**Describe the bug**

Nothing validates that scene ids within a `Task` are unique, but downstream code
derives per-trial identity from them. `rollout.py:247` builds
`trial_id = f"{scene.id}-e{epoch}"`, and `FrameStore.put` turns that into a
filename (`frames.py:59`). Two scenes sharing an id therefore write to the same
frame paths, and the second silently overwrites the first: half the run's stored
frames disappear with no warning, no error, and a log that still claims both
trials ran.

The same assumption is made in `_summarize.py:168-170`, which keys transcripts by
`(scene_id, epoch)`:

```python
transcript_by_key = {
    (transcript.scene_id, transcript.epoch): transcript for transcript in transcripts
}
```

so one duplicate scene's transcript is dropped and the surviving one is reported
twice under both rows.

I'd file this against `Task` rather than either consumer: both consumers are
reasonable to assume unique ids, and patching them individually leaves the next
consumer to rediscover it.

**To reproduce**

A/B, identical except for the scene ids:

```python
import tempfile, os, glob
from inspect_robots import eval as ir_eval
from inspect_robots.task import Task
from inspect_robots.scene import Scene
from inspect_robots.mock import CubePickEmbodiment, ScriptedPolicy
from inspect_robots.scorer import success_at_end

def run(ids, label):
    tmp = tempfile.mkdtemp()
    task = Task(
        name="t",
        scenes=[Scene(id=i, instruction="reach the cube", init_seed=n)
                for n, i in enumerate(ids)],
        scorer=success_at_end(), max_steps=5,
    )
    log = ir_eval(task, policy=ScriptedPolicy(), embodiment=CubePickEmbodiment(),
                  log_dir=tmp, store_frames=True)[0]
    files = sorted(os.path.basename(p)
                   for p in glob.glob(os.path.join(log.stats.frames_dir, "*.npy")))
    prefixes = sorted({f.rsplit('_', 2)[0] for f in files})
    print(f"{label}: scene ids={ids}")
    print(f"   trials={log.results.total_trials}  frame_files={len(files)}  trial_prefixes={prefixes}")
    return len(files)

a = run(["s1", "s2"], "UNIQUE ids  ")
b = run(["same", "same"], "DUPLICATE ids")
print(f"\nframes written with unique ids:    {a}")
print(f"frames written with duplicate ids: {b}")
print(f"SILENTLY LOST: {a - b} frame files")
```

```
UNIQUE ids  : scene ids=['s1', 's2']
   trials=2  frame_files=12  trial_prefixes=['s1-e0', 's2-e0']
DUPLICATE ids: scene ids=['same', 'same']
   trials=2  frame_files=6  trial_prefixes=['same-e0']

frames written with unique ids:    12
frames written with duplicate ids: 6
SILENTLY LOST: 6 frame files
```

The log reports `total_trials: 2` and two samples, both with
`scene_id='same'`, so nothing downstream can tell the two trials apart or notice
that one trial's frames are gone.

**Expected behavior**

`Task` should reject duplicate scene ids at construction with a `ConfigError`
naming the offending id, the way it already rejects every other configuration
that would make a rollout meaningless — `Task.__post_init__` raises `ConfigError`
for a missing/double horizon (`task.py:75`), a `max_steps` that is boolean or
`< 1` (`task.py:79`), a non-finite or non-positive `max_seconds` (`task.py:84`),
and `Epochs.count < 1` (`task.py:32`). Given that level of upfront validation,
silently losing half a run's frames to a duplicate id reads as an oversight
rather than a deliberate permission.

If duplicate ids are meant to be legal, then the frame path and the digest key
both need to incorporate something else that is actually unique (the scene's
index in `task.scenes`), which is a bigger change and would alter existing frame
filenames.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

**Additional context**

Happy to send the fix: a uniqueness check in `Task.__post_init__` raising
`ConfigError` with the duplicate id named, plus a regression test. That is one
guard in one place and leaves both consumers untouched.

Rishet11:
Fixes #289.

## What was wrong

Nothing validated that scene ids in a `Task` are unique, but downstream code
derives per-trial identity from them: `rollout.py:247` builds
`trial_id = f"{scene.id}-e{epoch}"`, and `FrameStore.put` turns that into a
filename (`frames.py:59`). Two scenes sharing an id wrote to the same frame
paths, so the second silently overwrote the first — half the run's stored frames
gone, no warning, and a log still reporting both trials with indistinguishable
`scene_id`s.

A/B on `main`, identical apart from the ids:

```
UNIQUE ids  : scene ids=['s1', 's2']
   trials=2  frame_files=12  trial_prefixes=['s1-e0', 's2-e0']
DUPLICATE ids: scene ids=['same', 'same']
   trials=2  frame_files=6  trial_prefixes=['same-e0']
```

`_summarize.py:168-170` keys transcripts by `(scene_id, epoch)` on the same
assumption, so one duplicate's transcript is dropped and the survivor is
reported twice.

## The fix

Validate uniqueness in `Task.__post_init__`, raising `ConfigError` with the
duplicate id named. That sits alongside the guards already there for the
horizon (`task.py:75`), `max_steps` (`task.py:79`), `max_seconds` (`task.py:84`)
and `Epochs.count` (`task.py:32`).

I fixed it at the `Task` boundary rather than in either consumer on purpose: both
consumers are reasonable to assume unique ids, and patching them one at a time
leaves the next consumer of `trial_id` to rediscover the same thing. It also
fails at construction, before a long run does work it will silently lose.

**This rejects input that previously ran**, so it is a behaviour change, not
purely additive. A task with duplicate ids was already losing data, so I do not
think anyone can be relying on it, but flagging it explicitly since it will turn
a silently-degraded run into a startup error.

## Tests

`test_task_rejects_duplicate_scene_ids` asserts the `ConfigError` and that the
message names the offending id. Against unfixed code it fails with
`Failed: DID NOT RAISE ConfigError`.

No existing test relied on duplicate scene ids: the suite goes from 1136 to 1137
passing with only the new test added.

## Verification

```
```

Test-merged against my other open branches (#250, #251, #256, #286, #288): no
conflicts, CHANGELOG entry deliberately anchored away from theirs.

- **No change to `FrameStore` or the digest key.** Making frame paths unique by
  scene index would work too, but it would change existing frame filenames and
  leaves the underlying "ids are identity" assumption unstated. If you would
  rather duplicate ids stay legal, this PR is the wrong approach and I would
  redo it that way instead.
- **No uniqueness check on anything else** (task names across an eval-set,
  camera names, dim labels). Out of scope for #289.
- **No epoch involvement.** `(scene_id, epoch)` is already unique once scene ids
  are, so the fix is one guard rather than a key change.


jeqcho (approved):
Thanks @Rishet11 — this is a model bug report and fix. The issue's A/B repro made the data loss concrete, and the PR description pre-answered every question a reviewer would ask.

Verified during review:

- **Bug is real.** `src/inspect_robots/rollout.py:264` builds `trial_id = f"{scene.id}-e{epoch}"`, `FrameStore.put` (`src/inspect_robots/frames.py:59`) derives frame filenames from it, and `src/inspect_robots/_summarize.py:169` keys transcripts by `(scene_id, epoch)` — duplicate ids silently overwrite frames and drop a transcript, exactly as described.
- **Right layer.** Validating in `Task.__post_init__` matches the existing horizon/`max_steps`/`max_seconds`/`Epochs` guards there, keeps both consumers untouched, and fails fast at construction instead of mid-run. Agreed that patching consumers individually would just defer the problem.
- **No sanitization loophole.** I checked whether two distinct ids could still collide after filename sanitization; `_safe` (`src/inspect_robots/frames.py:24`) appends a hash when it rewrites characters precisely to prevent that, so raw-id uniqueness is the complete guard.
- **Test placement and style fit** — the regression test sits beside the other `Task` validation tests in `tests/test_types_spaces.py`, asserts both the exception type and that the message names the offending id, and the error message format matches its neighbors.
- The behaviour change (previously-running input now rejected) is correctly flagged, and since duplicate-id runs were already silently losing data, failing loudly is the right trade.

One optional nit, fine to leave as-is: when several ids are duplicated, the error names only the first one hit, so a user fixes them one construction attempt at a time. Collecting all duplicates into one message would be a small polish, not a blocker.

jeqcho (approved):

PR #293, opened 2026-08-04
fix(logging): cap the log filename slug so a long task name still writes

### LINKED ISSUE #292: [bug] A long task name makes JsonLogSink fail to write, losing the entire run's log
**Describe the bug**

`JsonLogSink.on_eval_end` derives the log filename from the task name with no
length cap (`logging/json_log.py:86`):

```python
filename = f"{_slug(log.eval.task)}_{uuid.uuid4().hex[:8]}.json"
```

and `_slug` only substitutes characters, never truncates
(`logging/json_log.py:34-35`). A task name long enough to push the filename past
the filesystem's 255-byte limit makes the write raise `OSError: [Errno 63] File
name too long`.

The damage is that this happens in `on_eval_end`, after every trial has run. The
temp file cannot be created, so nothing is written and nothing is recovered: the
log directory is left empty and the entire run's results are gone. A long run
(hardware time, API spend) is lost at the final step, for a reason that has
nothing to do with the eval.

Threshold on macOS/APFS: the temp path is
`slug + "_" + 8 hex + ".json.tmp"`, i.e. `len(slug) + 18`, so it fails at roughly
238 characters of task name.

**To reproduce**

```python
import tempfile, os
from inspect_robots import eval as ir_eval
from inspect_robots.task import Task
from inspect_robots.scene import Scene
from inspect_robots.mock import CubePickEmbodiment, ScriptedPolicy
from inspect_robots.scorer import success_at_end

for n in (200, 240):
    tmp = tempfile.mkdtemp()
    task = Task(name="a" * n, scenes=[Scene(id="s", instruction="reach the cube", init_seed=0)],
                scorer=success_at_end(), max_steps=3)
    try:
        ir_eval(task, policy=ScriptedPolicy(), embodiment=CubePickEmbodiment(), log_dir=tmp)
        print(f"name len={n} -> OK")
    except Exception as e:
        print(f"name len={n} -> {type(e).__name__}: {e}")
        print(f"   log_dir contents after failure: {sorted(os.listdir(tmp))}")
```

```
name len=200 -> OK
name len=240 -> OSError: [Errno 63] File name too long: '/var/folders/.../aaaa...aaa_4a76401d.json.tmp'
   log_dir contents after failure: []
```

The empty `log_dir` is the point: the run finished, and its results are
unrecoverable.

**Expected behavior**

The sink should write the log regardless of task-name length, capping the slug so
the derived filename stays within the filesystem limit. Uniqueness is already
handled by the `uuid4().hex[:8]` component, so truncation cannot collide two logs.

There is in-repo precedent for treating a derived filename as something to be
made safe rather than trusted: `FrameStore` runs names through `frames._safe`
(`frames.py:24-33`), which rewrites unsafe characters and appends a CRC when it
had to change anything, specifically so distinct inputs stay distinct. The JSON
sink does the character half of that (`_slug`) but not the length half.

A guided `ConfigError` at `Task` construction would be a defensible alternative,
but capping seems better: the task name is display metadata, and a name being
long is not itself a configuration error.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64), APFS
- NumPy version: 2.4.6

**Additional context**

The task name is preserved in full inside the log (`log.eval.task`), so capping
the filename loses nothing — `inspect`/`view` read the name from the JSON, not the
path. Happy to send the fix.

Rishet11:
Fixes #292.

## What was wrong

`JsonLogSink.on_eval_end` derives the log filename from the task name
(`logging/json_log.py:86`) and `_slug` only substituted characters, never
truncated. A task name past roughly 238 characters pushed the temp path
(`slug` + `_` + 8 hex + `.json.tmp`) over the 255-byte limit.

Because `on_eval_end` runs after every trial has completed, the resulting
`OSError: [Errno 63] File name too long` cost the entire finished run: the temp
file could not be created, nothing was written, nothing recovered, and the log
directory was left empty.

```
name len=200 -> OK
name len=240 -> OSError: [Errno 63] File name too long: '.../aaa...aaa_4a76401d.json.tmp'
   log_dir contents after failure: []
```

That empty directory is the whole problem — hardware time or API spend, then no
results, for a reason unrelated to the eval.

## The fix

Cap the slug at 200 characters, leaving room for the suffix with headroom for
filesystems that allow less than 255. Two details that make capping safe:

- **Uniqueness is unaffected.** The filename already carries `uuid4().hex[:8]`,
  so two long names truncating to the same slug still get distinct files.
- **Nothing is lost.** The task name is recorded in full in the log body
  (`log.eval.task`), and `inspect`/`view` read it from the JSON rather than the
  path — the test asserts the full 300-character name round-trips.

I also re-`strip("-")` after truncating so a cut landing mid-separator does not
leave a trailing dash.

I chose capping over rejecting long names at `Task` construction: the task name is
display metadata, and a long name is not itself a configuration error. If you'd
rather it were a `ConfigError`, that's a small change and I'll redo it.

## Tests

`test_long_task_name_still_writes_its_log` runs a real eval with a 300-character
task name, asserts the written filename is within 255 bytes, and asserts the full
name survives in both the returned log and the re-read one. Against unfixed code
it fails with the production error:

```
OSError: [Errno 63] File name too long: '.../aaa...aaa_63c91fa3.json.tmp'
```

## Verification

```
```

Test-merged against my other open branches (#250, #251, #256, #286, #288, #290):
no conflicts.

- **No cap on the frames directory or any other derived path.**
  `FrameStore` builds its own names through `frames._safe`, which has the same
  no-length-cap property; I did not touch it because I have not reproduced a
  failure there and it is a separate path. Worth a look if you want it chased.
- **No validation of task-name characters.** `_slug` already handles that.
- **No change to the atomic write.** `os.replace` behaviour is fine; the failure
  was purely the name.


jeqcho (approved):
Thanks @Rishet11 — this is a well-argued fix for a genuinely painful failure mode (losing a finished run's log at the last step), and both the issue and PR write-ups made verification easy.

What I checked:

- **The bug reproduces from the code as claimed.** `_slug` (`src/inspect_robots/logging/json_log.py:35`) only substituted characters, and the filename at line 86 adds 18 bytes (`_` + 8 hex + `.json.tmp`), so a ~238-char task name crosses 255 bytes exactly as described in #292.
- **The cap is safe.** Post-regex the slug is ASCII-only (`[a-z0-9-]`), so `[:200]` chars == 200 bytes, leaving a worst-case 218-byte filename with real headroom. The re-`strip("-")` after truncation is a nice touch, and the `or "eval"` fallback still behaves correctly (the re-strip can never empty a non-empty slug, since the first character is never a dash after the initial strip).
- **Uniqueness and data preservation hold.** The `uuid4().hex[:8]` suffix carries uniqueness, and the test proves the full 300-char name round-trips through `log.eval.task` and `read_eval_log`.
- **The test would catch a regression** — it runs a real eval end to end and asserts the written filename's byte length, matching the style of the surrounding tests in `tests/test_strict_json.py`.

One bonus worth noting for the record: because `live_log.py:112` and `rerun_sink.py:662` import `_slug` from `json_log`, this cap also fixes the same latent overflow in the `.live.json` and `.rrd` filenames (worst cases 219 and 213 bytes respectively) — so the fix covers more than the JSON sink for free.

Agree with leaving `frames._safe` for a separate issue, and with capping over a `ConfigError` — a long display name isn't a configuration error. LGTM.

jeqcho (approved):

PR #295, opened 2026-08-04
fix(agent): accept either quote when reading a camera label

### LINKED ISSUE #294: [bug] Agent plugin re-sends a camera frame when the camera name contains an apostrophe
**Describe the bug**

`LLMAgentPolicy` recovers which cameras are still present in context by
re-parsing its own rendered label text, and the parse assumes `repr()` used
single quotes (`policy.py:912-915`):

```python
if text.startswith("camera '"):
    end_quote = text.find("'", 8)
    if end_quote != -1:
        active_camera_names.add(text[8:end_quote])
```

The label it is parsing is built with `!r` (`policy.py:1167`):

```python
f"camera {name!r}{suffix}:"
```

Python's `repr()` switches to double quotes when the string contains a `'` and no
`"`, so a camera named `front's cam` renders as `camera "front's cam":`. The
`startswith("camera '")` test then fails, the camera is never added to
`active_camera_names`, and `self._revealed.intersection_update(active_camera_names)`
(`policy.py:916`) drops it from the revealed set even though its image is still in
context.

The consequence is that the "already shown for this observation" guard
(`policy.py:918-936`) stops firing for that camera: a second `take_pic` in the
same observation turn re-captures and re-sends the identical frame. That
contradicts the guarantee the system prompt makes to the model
(`policy.py:118-119`) and the rejection path at `policy.py:933-937`, bills image
tokens twice, and leaves a transcript implying the model saw a fresh view when
nothing moved.

The hardcoded `8` (the length of `"camera '"`) encodes the same single-quote
assumption.

**To reproduce**

With a stubbed `httpx.MockTransport` (no network), one camera, `images="on_demand"`,
`image_horizon=1`, and a script of `take_pic` → `take_pic` (same camera) →
`give_up`:

Camera named `front's cam`:

```
image-bearing message count (sanitized): 2
tool results in order:
 - captured 1 frame(s): "front's cam"
 - captured 1 frame(s): "front's cam"
 - give_up: test done
```

Identical script, camera renamed `front_cam`:

```
image-bearing message count (sanitized): 1
tool results in order:
 - captured 1 frame(s): 'front_cam'
 - already shown for this observation: 'front_cam'; the view cannot change until the robot moves
 - give_up: test done
```

The mechanism on its own:

```python
>>> repr("front's cam")
'"front\'s cam"'
>>> ("camera " + repr("front's cam") + ":").startswith("camera '")
False
```

**Expected behavior**

The second `take_pic` for the same camera in the same observation should be
rejected with the existing "already shown for this observation … the view cannot
change until the robot moves" result, exactly as it is for a camera whose name
contains no apostrophe. A camera's name should not change the policy's
image-dedup behaviour.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

**Additional context**

Narrow trigger — the name needs a `'` and no `"` — but it is silent and hits every
turn for an embodiment that names a camera something like `robot's wrist`. Every
camera name in the existing tests is a plain identifier, which is why this is
uncovered.

Happy to send the minimal fix (accept either quote character and derive the offset
from the prefix instead of hardcoding `8`). The sturdier fix is to stop recovering
structured state by re-parsing rendered prompt text at all and track the revealed
cameras per message directly — larger, and your call whether it's worth it.

Rishet11:
Fixes #294.

## What was wrong

`LLMAgentPolicy` recovers which cameras are still in context by re-parsing its own
rendered label, and the parse assumed `repr()` used single quotes
(`policy.py:912-915`):

```python
if text.startswith("camera '"):
    end_quote = text.find("'", 8)
```

The label is written with `!r`. `repr()` switches to double quotes for a string
containing a `'` and no `"`, so a camera named `robot's wrist` renders as
`camera "robot's wrist":`, the `startswith` test fails, the name never enters
`active_camera_names`, and `intersection_update` then empties the revealed set
even though the frame is still in context.

The visible effect is that the "already shown for this observation" guard stops
firing: a second `take_pic` for the same camera in the same observation re-sends
the identical frame. That contradicts what the system prompt promises the model
(`policy.py:118-119`), bills image tokens twice, and leaves a transcript implying
a fresh view when nothing moved.

## The fix

Accept whichever quote `repr()` chose, and take the offset from a
`_CAMERA_LABEL_PREFIX` constant that the writer now uses too. The hardcoded `8`
was the length of `"camera '"` — the same single-quote assumption, spelled twice.

Sharing the prefix between writer and reader is the part I'd argue for: the bug is
a producer/consumer skew, and one constant is what stops it recurring. The
rendered output is byte-identical (`"camera " + repr(name) + suffix`), so the
existing byte-stability test still passes unchanged.

## Tests

`test_on_demand_dedups_a_camera_whose_name_contains_an_apostrophe` runs the real
`act()` loop over a stubbed `httpx.MockTransport` with one camera named
`robot's wrist` and a `take_pic` → `take_pic` → `move_by` script, asserting the
second capture is refused.

Against unfixed code it fails with the defect itself:

```
assert 'captured 1 frame(s): "robot\'s wrist"'.startswith('already shown for this observation:')
E   assert False
```

## Verification

```
pytest --cov (core)          1136 passed, 5 skipped   Total coverage: 100.00%
pytest plugins/...-agent     470 passed
ruff format                  127 files already formatted
mypy                         Success: no issues found in 74 source files
ruff check                   2 errors, both pre-existing in this same file
                             (policy.py:286,291 union order, shifted by 3 lines
                             from 283,288 by my added constant) and untouched
```

Note the core `--cov` run does not cover plugin code — coverage source is scoped
to `inspect_robots` — so the plugin suite is the gate that exercises this change,
matching how CI runs it as its own job.

## CHANGELOG conflict, deliberately

This entry sits at the same anchor as my #288, so **this branch and #288 conflict
in CHANGELOG.md and nothing else**. That was a choice: the `### Fixed` section's
insertion points are all currently claimed by open PRs, so rather than collide
with one of the approved-and-waiting ones (#250, #251, #256), I put the collision
on my own newest PR. Verified against every other open branch of mine (#250, #251,
#256, #286, #290, #293): no conflicts.

Merge either this or #288 in any order and I'll rebase the other immediately — or
tell me where you'd rather the entry go and I'll move it.

- **No move away from parsing rendered prompt text.** Tracking revealed cameras
  structurally per message would remove this whole class of bug, but it is a real
  refactor of the eviction bookkeeping and not something I'd slip into a bug fix.
  Happy to write it up separately if you want it.
- **No change to the label format.** Quoting names with `!r` is fine; only the
  reader was wrong, and changing the format would churn the wire captures.
- **No plugin version bump.** Existing entries use both versioned and unversioned
  "Agent plugin:" prefixes; I used the unversioned form since the release
  cadence is yours to decide.


jeqcho (approved):
Thanks for another meticulous fix, @Rishet11 — the issue write-up (#294) made this trivially verifiable, and the fix is exactly the right size.

**What I checked:**

- **The bug is real.** On current main, `policy.py:954` does `text.startswith("camera '")` while the writer at `policy.py:1211` renders with `{name!r}` — and `repr()` switches to double quotes for a name containing an apostrophe. The reproduction in the issue is accurate: the label never matches, `intersection_update` empties `_revealed`, and the "already shown for this observation" guard goes dead for that camera.
- **The fix is correct.** The reader now derives its offset from the shared `_CAMERA_LABEL_PREFIX` and accepts whichever quote `repr()` chose; the short-text case is safely guarded (`quote in ("'", '"')` fails on an empty slice). The writer change is byte-identical output, so nothing in the wire captures or the byte-stability test moves. Sharing one constant between producer and consumer is the right structural remedy for a producer/consumer skew — the hardcoded `8` was the same assumption spelled twice.
- **The test earns its keep.** It drives the real `act()` loop through `httpx.MockTransport` with a `take_pic` → `take_pic` → `move_by` script, and the expected strings match the actual `_quoted_names` (repr-based) formatting, so it fails on unfixed code with the defect itself rather than an incidental assertion. The in-function `numpy`/`Observation` imports mirror the existing sibling test at `test_policy_eviction.py:471-473`, so it's consistent with the file's conventions.
- **CI:** all 18 checks green, including `plugin · agent`, which is the job that actually exercises this code given the core `--cov` scoping you noted.

**One non-blocking observation:** a name containing *both* quote characters (or a backslash) still won't round-trip — `repr()` then backslash-escapes, and the reader would extract the escaped form, which won't match the real name in `_revealed`. That's the same pre-existing fragility class you called out, and I agree it belongs in the structural "track revealed cameras per message" follow-up rather than this fix. If you're willing to file that as an issue, I'd welcome it.

Also appreciated the deliberate CHANGELOG conflict management with #288 — placing the collision on your own newest PR was the considerate call. Merging this one.

jeqcho (approved):

PR #297, opened 2026-08-04
fix(controller): give each SmoothingController its own EMA state

### LINKED ISSUE #296: [bug] Two SmoothingControllers in one trial share EMA state and corrupt each other
**Describe the bug**

`SmoothingController` keeps its EMA state in the per-trial store under a single
module-level key (`controller.py:69`):

```python
_SMOOTH_KEY = "_smoothing_prev_action"
```

and reads/writes it unconditionally (`controller.py:93-95`):

```python
prev = store.get(_SMOOTH_KEY)
smoothed = raw if prev is None else self.alpha * raw + (1 - self.alpha) * prev
store[_SMOOTH_KEY] = smoothed
```

Because the key is per-module rather than per-instance, two `SmoothingController`
instances active in the same trial share one slot and destructively overwrite each
other. Each layer then smooths against whatever the other last wrote, so both
produce silently wrong actions — no crash, no warning.

The class docstring presents exactly this composition as the point of the design
(`controller.py:74-79`): "Demonstrates the middleware composition the
single-method interface enables". Nothing warns against stacking two of them, e.g.
a coarse and a fine smoothing pass with different `alpha`.

**To reproduce**

```python
import numpy as np
from inspect_robots.controller import DefaultController, SmoothingController
from inspect_robots.policy import PolicyInfo
from inspect_robots.spaces import ActionSemantics, Box
from inspect_robots.types import Action, ActionChunk, Observation

SPACE = Box(shape=(1,), semantics=ActionSemantics(control_mode="joint_delta"))

class Feed:
    def __init__(self, vals):
        self.vals = list(vals); self.i = 0
        self.info = PolicyInfo(name="feed", action_space=SPACE)
    def act(self, obs):
        v = self.vals[self.i]; self.i += 1
        return ActionChunk(actions=[Action(data=np.array([v]))])

vals = [1.0, 5.0, 2.0, 8.0, 3.0]
obs = Observation(instruction="x")
outer = SmoothingController(SmoothingController(DefaultController(), alpha=0.5), alpha=0.9)
store: dict = {}
actual = [float(np.asarray(outer.next_action(Feed(vals[i:]), obs, i, store).data)[0])
          for i in range(len(vals))]

inner_prev = None; outer_prev = None; expected = []
for v in vals:
    inner = v if inner_prev is None else 0.5 * v + 0.5 * inner_prev
    inner_prev = inner
    out = inner if outer_prev is None else 0.9 * inner + 0.1 * outer_prev
    outer_prev = out
    expected.append(out)

print("shared store (actual)  :", [round(x, 4) for x in actual])
print("independent (expected) :", [round(x, 4) for x in expected])
print("keys in store:", [k for k in store if "smooth" in k])
```

```
shared store (actual)  : [1.0, 3.0, 2.5, 5.25, 4.125]
independent (expected) : [1.0, 2.8, 2.53, 4.978, 4.2103]
keys in store: ['_smoothing_prev_action']
```

One key for two controllers, and the streams diverge from the second step on.

**Expected behavior**

Each `SmoothingController` instance should keep its own EMA state, so stacking two
gives the mathematically correct nested EMA (the `expected` row above).

Note the contrast with the other controller state keys, which are safe to share for
a specific reason: `_INFER_KEY` is only ever **appended** to (`controller.py:65`,
`controller.py:159`), so `DefaultController` and `EnsemblingController` coexisting
just add entries. `_SMOOTH_KEY` is the only one written destructively, which is
what makes sharing it unsafe.

**Environment**
- Python version: 3.13.3
- OS: macOS 26.5.1 (arm64)
- NumPy version: 2.4.6

**Additional context**

Being straight about reachability: stacking two smoothing layers is an unusual
configuration, and I hit this by reading for state-isolation bugs rather than from a
real run. So this is latent rather than something users are tripping over today. I
think it is still worth fixing because it is silent, it produces wrong actuation
rather than an error, and the class is public and documented as composable.

Happy to send the fix: namespace the key per instance so the state cannot be
shared. If you would rather `SmoothingController` simply refuse to be nested, that
is also a clean answer and a smaller behavioural surface — say which you prefer.

Rishet11:
Fixes #296.

## What was wrong

`SmoothingController` kept its EMA state in the per-trial store under a single
module-level key (`controller.py:69`), read and written destructively
(`controller.py:93-95`). Two instances active in the same trial therefore shared
one slot: each layer smoothed against whatever the other last wrote, so both
emitted something that is neither layer's intended output — silently, with no
crash or warning.

The class docstring presents this exact composition as the point of the design
("Demonstrates the middleware composition the single-method interface enables"),
and nothing warns against stacking two.

Stacking `alpha=0.9` over `alpha=0.5` on the stream `[1, 5, 2, 8, 3]`:

```
shared store (actual)  : [1.0, 3.0, 2.5,   5.25,  4.125]
independent (expected) : [1.0, 2.8, 2.53,  4.978, 4.2103]
keys in store: ['_smoothing_prev_action']
```

One key for two controllers, diverging from the second step on.

## The fix

Namespace the key per instance via a module-local serial. The contrast that makes
this the right shape: `_INFER_KEY` is safe to share because it is only ever
**appended** to (`controller.py:65`, `controller.py:159`), so `DefaultController`
and `EnsemblingController` coexisting merely add entries. `_SMOOTH_KEY` was the
only destructively written controller state, which is exactly why sharing it broke.

I used a counter rather than `id(self)` so the key is deterministic and readable in
a dumped store.

## Tests

`test_stacked_smoothing_controllers_keep_separate_state` stacks two layers and
compares against the independently computed nested EMA. Against unfixed code:

```
assert np.allclose([1.0, 3.0, 2.5, 5.25, 4.125],
                   [1.0, 2.8000000000000003, 2.5300000000000002, 4.978000000000001, 4.2103])
E   assert False
```

Worth noting explicitly: my first attempt at this test **passed against the broken
code**, because I drove it with `ScriptedPolicy`, whose oracle emits near-constant
steps — with a constant stream the corruption is invisible. The test needs a
varying action stream, which is why it uses a small local stub policy. I mention it
because it is the kind of regression test that would otherwise look fine and assert
nothing.

## Verification

```
```

## CHANGELOG conflict, deliberately

The `### Fixed` insertion points are all currently claimed by open PRs, so this
entry shares an anchor with my #290 — **those two conflict in CHANGELOG.md and
nothing else**. I first placed it where it collided with #251, which is approved
and waiting, and moved it specifically so the collision lands on my own newer PR
instead. Verified clean against #250, #251, #256, #286, #288, #293, #295.

Merge #290 or this in any order and I'll rebase the other immediately.

- **No refusal to nest.** Making `SmoothingController` reject an inner
  `SmoothingController` would also close the bug and is a smaller behavioural
  surface; I went with isolating state because the docstring advertises
  composition. If you'd rather it refuse, say so and I'll swap it.
- **No change to `_INFER_KEY` or `_BUFFER_KEY`.** They are shared safely for the
  reason above, and touching them would be scope creep.
- **On reachability, plainly:** stacking two smoothers is unusual, and I found this
  by reading for state-isolation bugs, not from a failing run. It is latent. I
  think it earns a fix because it is silent and produces wrong actuation, but it is
  not urgent, and I would understand closing it as won't-fix.


jeqcho (approved):
Thanks for a really well-executed fix, and for the unusually honest PR write-up — the note about your first test passing against the broken code (constant-stream corruption being invisible) is exactly the kind of detail that saves the next person a wasted afternoon.

What I checked:

- **The bug is real.** `_SMOOTH_KEY` at `src/inspect_robots/controller.py:69` is a single module-level key, read and destructively written in `next_action` (lines 93–95 pre-fix), so two stacked `SmoothingController`s share one EMA slot. I reproduced the arithmetic in the issue (#296) against the code and it holds: from step 2 on, each layer smooths against the other's last write. Given the class docstring explicitly advertises middleware composition, fixing rather than forbidding nesting is the right call.
- **The fix is correct and minimal.** Per-instance key minted at construction (`f"{_SMOOTH_KEY}:{next(_smooth_serial)}"`) keeps state trial-local via the store rather than moving it onto `self` — important, since the store is created fresh per trial (`rollout.py:267`), so a reused controller instance still gets a clean EMA each trial. Storing `prev` on the instance would have leaked state across trials; this shape avoids that. I also confirmed the store is never persisted to `EvalLog`, so the process-local serial comment is accurate, and the counter-over-`id(self)` choice is sensible for readable dumped stores.
- **The test bites.** `test_stacked_smoothing_controllers_keep_separate_state` in `tests/test_controller.py` uses a varying action stream (chunk length 1, so `DefaultController` re-infers every step) and compares against an independently computed nested EMA — the failure output against unfixed code in the PR body confirms it would have caught the regression.
- **Hygiene.** All CI checks green including the 100% coverage gate, mypy strict over tests, and the branch is mergeable. The contrast note about `_INFER_KEY`/`_BUFFER_KEY` being append-only (and therefore safe to share) is a genuinely useful comment for future controller authors.

One non-blocking observation: a `copy.deepcopy`/pickle of a `SmoothingController` would duplicate its store key and reintroduce sharing between the copies. Nothing in the codebase does that today, so I don't think it needs handling — just flagging it in case controller cloning ever becomes a pattern.

Also appreciated the deliberate CHANGELOG anchor choice so the conflict lands on your own #290 rather than #251. LGTM.

jeqcho (approved):

PR #307, opened 2026-08-04
feat(wizard): support context-aware wizard default suggestions in Opt…

### LINKED ISSUE #303: OPTION_SLOTS: context-aware wizard suggestions (plugin sees the carried config when suggesting a default)
inspect-robots-yam#110 flipped the wizard's collision_guardrail suggestion to off because fresh rigs have no measured geometry. The right suggestion is actually conditional: a carried config that already sets `collision_left/right_base_pos` is a measured rig, and the wizard should suggest **on** there. `OptionSlot.default` is a static bool, so the plugin cannot express this — and the current behavior has a real downgrade path: a hand-written config with geometry keys but no explicit `collision_guardrail` key gets suggested off on a setup re-run, and an Enter-accept disables an active guardrail on a measured rig (documented as a caveat in yam's README for now).

Proposal: let an option slot compute its suggestion from the carried `[embodiment.args]` mapping — e.g. an optional `suggest(existing_args: Mapping[str, str]) -> bool` on the slot (or a core-passed context object), falling back to `default`. The existing carried-value override (an explicit stored bool wins) should keep top precedence.

Adityakk9031:
## Summary
Resolves #303.

This PR updates the embodiment setup wizard's `OptionSlot` schema to support context-aware default suggestions. Instead of relying solely on a static fallback boolean when a configuration key is absent, an embodiment plugin can now provide a dynamic `suggest` callback:

```python
suggest: Callable[[Mapping[str, str]], bool] | None = None
```

This enables plugins (such as `inspect-robots-yaml`) to recommend sensible defaults based on the existing configuration. For example, a plugin can automatically suggest enabling `collision_guardrail` when `collision_left/right_base_pos` is already configured in `[embodiment.args]`.

## Changes

- **`conformance.py`**
  - Added a `suggest` callback field to the `OptionSlot` dataclass.

- **`_setup.py`**
  - Updated `_options_section` to invoke `suggest`, passing the current `[embodiment.args]` mapping.
  - Explicit configuration values continue to take highest precedence over suggested defaults.

- **`test_setup.py` / `test_conformance.py`**
  - Added tests verifying that:
    - Context-aware suggestions are applied when appropriate.
    - Static fallback behavior is preserved when no context is available.
    - Explicitly configured values override suggested defaults.

- **`CHANGELOG.md`**
  - Documented support for context-aware suggestion callbacks.

## Verification

Ran the setup and conformance test suites successfully:

```bash
python -m pytest tests/test_conformance.py tests/test_setup.py -o pythonpath=src
```


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks for picking this up, @Adityakk9031 — this lands exactly the shape proposed in #303, and the implementation gets the important things right: the `suggest` result is computed before the carried-value check, so an explicit stored bool keeps top precedence, and the three tests in `tests/test_setup.py` (no-context fallback, context-aware flip to `[Y/n]`, explicit-config override) cover the precedence matrix nicely, including asserting on the rendered prompt suffix. Good call amending the existing unreleased `OptionSlot` changelog bullet rather than adding a new one.

A few things before this is ready to merge:

1. **Plugin-facing docs are missing** — `docs/guide/adapters.md` (the `OPTION_SLOTS` section, ~lines 103–117) documents the plugin contract that this PR extends, but it isn't updated. Since `suggest` is exactly the kind of thing an adapter author discovers through that guide, please add a short mention (signature plus a one-line example, e.g. the `collision_guardrail`-from-geometry case from the issue).

2. **`OptionSlot` docstring not updated** — the class docstring in `src/inspect_robots/conformance.py` describes each field (`arg`, `label`, `default`) but says nothing about `suggest`. A sentence noting it computes the suggestion from the carried `[embodiment.args]` and that its failures/absence fall back to `default` would keep the dataclass self-documenting, matching the style used elsewhere in this module.

3. **Non-bool returns from `suggest` flow through unvalidated** — in `_options_section` (`src/inspect_robots/_setup.py`), `contextlib.suppress(Exception)` guards against a raising callback (good, and consistent with this codebase's "a plugin bug must never crash the wizard" posture), but a callback returning a non-bool (say the string `"false"`, which is truthy) is assigned to `suggested` as-is and drives `_ask_yes_no` via truthiness — Enter would then write the wrong value. The sibling carried-value branch right below already guards with `isinstance(parsed, bool)`, and `option_slots` / `missing_runtime_requirements` both type-filter plugin data the same way. Suggest mirroring that:

   ```python
   if option.suggest is not None:
       try:
           computed = option.suggest(existing_args)
       except Exception:
           computed = None
       if isinstance(computed, bool):
           suggested = computed
   ```

   As a small optional hardening while you're there: `existing_args` is the live carried dict, so a buggy callback could mutate it; wrapping it in `types.MappingProxyType` (the declared type is already read-only `Mapping[str, str]`) would close that off cheaply.

None of these touch the core design, which looks right. Happy to re-review once the docs and the bool guard are in — thanks again for the clean tests!

Adityakk9031:

1. Added plugin-facing documentation for `OptionSlot.suggest` in `docs/guide/adapters.md`.
2. Updated `OptionSlot` class docstring in `src/inspect_robots/conformance.py` to explain the `suggest` callback and its fallback behavior.
3. Hardened `_options_section` in `src/inspect_robots/_setup.py` by validating that `suggest` returns a boolean, wrapping `existing_args` in `types.MappingProxyType` to prevent accidental mutations, and falling back safely if exceptions or non-bool values are returned. Added tests for this fallback behavior in `tests/test_setup.py`.
4. Rebased onto latest `main`.

Adityakk9031:
Fixed CI lint & typecheck issues (imported \Callable\ in \conformance.py\, import ordering in \_setup.py\, strict mypy annotation in \	est_setup.py\). All local ruff and mypy checks pass cleanly.

/review

Adityakk9031:

1. **Clean imports & typing:** Fixed missing `Callable` import in `src/inspect_robots/conformance.py`, sorted stdlib imports in `src/inspect_robots/_setup.py`, and resolved mypy typing annotations in `tests/test_setup.py`.
2. **CI status:** Verified all unit tests, `ruff check`, `ruff format`, and `mypy --strict` pass cleanly (`ci-ok pass`). Ready for your final review.

PR #310, opened 2026-08-05
feat(defaults): expose public init_dotenv helper under defaults (#301)

### LINKED ISSUE #301: expose a public dotenv helper so plugins can mirror the CLI's .env loading without importing inspect_robots._dotenv
inspect-robots-yam#108 (issue inspect-robots-yam#107) makes the yam health/holdcheck CLIs honor the CWD `.env` exactly like the core CLI, which requires calling `init_dotenv`. The only import path today is the private module:

```python
from inspect_robots._dotenv import init_dotenv
```

That is a top-level import in two console scripts: if core ever renames or moves the module, both scripts die at import time (even `--help`). The plugin pins core tightly and a weekly canary would catch it, but the clean fix is a public re-export (e.g. `inspect_robots.dotenv.init_dotenv` or exporting from `inspect_robots.defaults` next to `config_path`, which plugins already import). Signature has been stable from 0.31 through 0.38.

Suggested by the fresh-eye review on inspect-robots-yam#108; the plugin should switch to the public path once it ships.

Adityakk9031:
## Description

Resolves #301.

Exposes `init_dotenv` as part of the public `inspect_robots.defaults` module API (next to `config_path`). This permits embodiment plugins (such as `inspect-robots-yam`) to mirror the core CLI's `.env` configuration file loading in health/holdcheck scripts without importing private/internal modules (`inspect_robots._dotenv`).

### Changes
* **defaults.py**: Imported `init_dotenv` from `inspect_robots._dotenv` and added it to the public `__all__` exports array.
* **test_defaults.py**: Added a test asserting that `inspect_robots.defaults` successfully exports `init_dotenv`.
* **CHANGELOG.md**: Added a changelog entry detailing the new publicdefaults helper.

### Verification
Ran and verified the defaults and dotenv tests successfully:
```bash
python -m pytest tests/test_defaults.py tests/test_dotenv.py -o pythonpath=src


Adityakk9031:
@jeqcho have a look

jeqcho (approved):
Thanks @Adityakk9031 for the clean, minimal implementation — this is exactly the shape #301 asked for (`inspect-robots-yam`'s health/holdcheck scripts currently import the private `inspect_robots._dotenv` path at module top level, so a public re-export is genuinely needed).

What I checked:

- **Correctness** — `src/inspect_robots/defaults.py` re-exports `init_dotenv` from `inspect_robots._dotenv` and adds it to `__all__` (kept alphabetized). `_dotenv` has no package-internal imports, so no circular-import risk from `defaults` importing it.
- **Surface discipline** — only `init_dotenv` is exported, not `read_dotenv`; matches the issue's scope and keeps the new public surface minimal. Placing it in `defaults` next to `config_path` (which plugins already import) was one of the two options the issue suggested.
- **Tests** — the exact-surface guard in `tests/test_defaults_public.py` (`test_public_surface_is_exact_and_resolves`) was updated, which is the load-bearing check: it asserts the export is present and resolves. CI is fully green across all matrices.
- **CHANGELOG** — entry lands under `[Unreleased] > Added` correctly.

A few optional nits, none blocking:

1. `tests/test_defaults.py:300` — `assert init_dotenv is not None` is a bit weak (the import alone already proves presence, and the exact-surface test covers it too). If you want it to earn its keep, assert identity instead: `assert defaults.init_dotenv is inspect_robots._dotenv.init_dotenv`, which pins that this is a re-export rather than a shadowing definition.
2. `CHANGELOG.md:46` — the entry is one ~150-char line in a file that wraps at ~80 columns, and neighboring entries link issues as `[#301](…)` rather than bare `#301`. Also conventional to add new entries at the top of the section rather than mid-list.
3. `src/inspect_robots/defaults.py:1` — the module docstring frames the module as config.ini reading; a one-line mention that it also re-exports the `.env` bootstrap helper would keep the docstring honest. Fine to leave for a follow-up.

Happy to merge as-is — thanks again for the tidy contribution, including remembering the exact-surface test.

jeqcho (approved):

PR #315, opened 2026-08-06
agent: fractional -P effort= for servers that read effort as a number

### LINKED ISSUE #314: agent: -P effort= should accept a fractional value, not only the named levels
`-P effort=` is validated against a frozenset of named levels
(`policy.py:_EFFORT_LEVELS`), so a number is rejected at construction:

```
effort must be one of ['high', 'low', 'max', 'medium', 'minimal', 'none', 'xhigh'], or None to omit the field, got 0.7
```

Some servers read reasoning effort as a continuous value instead. Probed
2026-08-06 against Tinker with `thinkingmachines/Inkling`:

| endpoint | `reasoning_effort: 0.7` | `1.0` / `0.995` | `"high"` |
|---|---|---|---|
| `.../anthropic/api/v1` (Messages) | 422, levels only | — | 200 |
| `.../oai/api/v1` (chat) | 200 | 422 | 200 |

So the accepted fractional range on Tinker's OpenAI-compatible endpoint is
`0.0` through `0.99`. The client should not be the thing standing in the way:
accept a number, pass it through unquantized, and let a server that takes
levels only say so in a guided 4xx.

Known limitation to document rather than solve here: the Tinker endpoint that
accepts a fraction silently ignores `tools`, so it cannot currently drive a
robot episode (the policy fails after three tool-call-free turns). Fractional
effort is therefore useful for prompt-level experiments now, and for real
rollouts when the Messages endpoint accepts a number.

jeqcho:
Closes #314.

## What

`-P effort=` now takes a number in `[0.0, 1.0)` in addition to the named
levels. The number is normalized to `float` and sent unquantized, so an effort
sweep keeps whatever resolution the server offers:

```bash
inspect-robots run --policy agent -P model=thinkingmachines/Inkling -P effort=0.7 ...
```

Named levels are unchanged and stay the portable choice. Validation now lives in
one `_validated_effort()` helper instead of a set-membership test inline in the
constructor.

## Why a number at all

Probed against Tinker on 2026-08-06 with `thinkingmachines/Inkling`:

| endpoint | `0.7` | `0.995` / `1.0` | `"high"` |
|---|---|---|---|
| `.../anthropic/api/v1` (Messages) | 422, levels only | — | 200 |
| `.../oai/api/v1` (chat) | 200 | 422 | 200 |

The accepted fractional range there is `0.0` through `0.99`. The upper bound
here is `1.0` rather than `0.99` on purpose: `1.0` is past the top of the range
on every server probed, while a provider-specific cap below it belongs to the
provider, which reports it as a 4xx we already guide on.

## Honest limitation

A fraction cannot drive a real rollout on Inkling today. The Messages endpoint
that serves it rejects a number, and the OpenAI-compatible endpoint that accepts
one silently ignores `tools`, so the policy sees no tool call and fails after
three turns. This PR removes the client-side blocker and documents that state in
the plugin README, so a fraction is usable for prompt-level experiments now and
for rollouts if or when the Messages endpoint takes one.

## Details

- Guided 4xx on the Messages wire distinguishes the two cases: a rejected
  fraction now names `-P wire=chat` and the endpoint that reads one, while a
  rejected `none`/`minimal` keeps the existing OpenAI-only-values message.
- `bool` is not treated as a number: `-P effort=false` parses to `False` and
  stays a `ConfigError` instead of silently meaning zero effort. Same treatment
  `image_horizon` already gives `True`.
- `nan` and `inf` fall out of the range check without a special case.
- The four wire clients widen `reasoning_effort` to `str | float | None`;
  `AgentPolicyConfig.effort` likewise, so the value round-trips into the log.
- Not in scope: the capx policy keeps its own level-only effort validation. It
  imports only public names from this plugin, so sharing the check would mean
  either duplicating it or widening a public API, neither of which belongs in
  this change.

## Test

Agent plugin suite: 481 passed. `ruff check`, `ruff format --check`, and strict
`mypy` clean. New coverage: fraction reaches the wire and the log unquantized,
`effort=0` sends `0.0` rather than omitting the field, each of `1.0`, `-0.1`,
`nan`, `inf`, `True`, `False`, and `"0.7"` raises, and the Messages-wire
fractional 4xx guidance fires. `uv.lock` refreshed for the plugin version bump
to 0.23.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ All notable changes to this project are documented here. The format is based on
 
 ### Added
 
+- **Agent plugin (0.24.0):** `-P effort=` now also takes a number in
+  `[0.0, 1.0)` for servers that read reasoning effort as a fraction rather than
+  a named level, and sends it unquantized so an effort sweep keeps the
+  resolution the server offers. Named levels and the 0.23.0 passthrough rules
+  are unchanged: an omitted flag still omits the field, and `none` still sends
+  the true minimum. Tinker's OpenAI-compatible endpoint accepts `0.0` through
+  `0.99`; wires that take levels only reject a fraction with a guided 4xx naming
+  the wire that accepts one. `-P effort=false` stays an error rather than
+  becoming zero effort (#314).
+
 - **Core (0.44.0):** operator messages now preserve console or attached-input
   provenance through transcripts, policy observations, and evaluation logs.
   `OperatorSession.attach_input()` merges feedback-only sources without risking
--- a/plugins/inspect-robots-agent/README.md
+++ b/plugins/inspect-robots-agent/README.md
@@ -325,6 +325,14 @@ level; omit the argument to inherit the provider default. Gemini Live has no
 effort field and rejects any explicit effort, so leave it unset on that wire.
 To pin the behavior from before version 0.23, add `-P effort=low`.
 
+Effort also takes a number in `[0.0, 1.0)` for servers that read it as a
+fraction instead of a named level (`-P effort=0.7`). The number is sent
+unquantized, so an effort sweep keeps whatever resolution the server offers.
+Named levels stay the portable choice: every wire and provider accepts some of
+them, while fractional effort is accepted today only by Tinker's
+OpenAI-compatible endpoint (see below). A server that takes levels only rejects
+a fraction with a guided 4xx naming the wire that does accept one.
+
 ## Depth rendering
 
 For each camera, the policy looks for metric depth in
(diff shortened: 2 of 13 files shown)


jeqcho:
Heads-up: #318 (effort passthrough, plan 0049) merged and took agent-plugin version **0.23.0**, touching the same `policy.py` effort-validation block. This branch needs a rebase: unset `effort` now resolves to `None` (field omitted → provider default, no more implicit `"low"`), a received Python `None` normalizes to the `"none"` level, and on `wire=messages` `"none"` maps to `thinking: {"type": "disabled"}`. Fractional values will want a stance on the messages wire too (Tinker's Messages endpoint is enum-only per the measurements in this PR). Re-bump to 0.24.0 (pyproject + uv.lock + test_package pin).

Issue #317, opened 2026-08-06
agent: -P effort=none silently means "omit the field", and unset effort injects "low" instead of the provider default


jeqcho:
## Two related surprises in `-P effort=` handling

**1. `-P effort=none` does not send effort "none" — it silently gets the provider default.**

The CLI's scalar parser (`_parse_value`, `src/inspect_robots/defaults.py`) coerces the literal `none`/`null` to Python `None` before the policy sees it. The agent policy defines `effort=None` as "omit the field from the request", so the provider applies its own server-side default. An operator typing `-P effort=none` always means the effort *level* `"none"` (it's in `_EFFORT_LEVELS`), but gets medium-ish provider-default reasoning instead — observed on both Inkling and Anthropic models. The only way to send the real value today is the quoting escape hatch `-P effort="'none'"`, which nobody discovers unaided.

**2. Omitting `-P effort=` doesn't mean "provider default" — the policy injects `"low"`.**

Unset effort resolves to an explicit `"low"` on every wire except gemini-live (`policy.py`). This contradicts operator intuition ("flags you don't set aren't sent") and is inconsistent with `temperature`, which the same policy omits when unset.

**3. There is no true "none" for Anthropic models at all.**

The Messages API rejects `effort: "none"` (OpenAI-only value), and the Messages shim hardcodes `thinking: {"type": "adaptive"}`. The faithful mapping for `effort=none` on `wire=messages` is `thinking: {"type": "disabled"}` — safe because effort is fixed per run, so no signed-thinking-block replay conflicts arise.

## Proposed behavior

| Operator input | Wire |
|---|---|
| *(flag omitted)* | effort/reasoning field omitted → provider default (matches `temperature`) |
| `-P effort=none` | `reasoning_effort: "none"` (chat), `reasoning.effort: "none"` (responses), `thinking: {"type": "disabled"}` (messages) |
| `-P effort=low`…`max` | unchanged |

- Constructor: received `None` normalizes to the string `"none"` (Python `None` can only arrive via the CLI coercion, and the operator always means the level); `_UNSET` → omit.
- gemini-live: any explicit effort (including `none`) stays a guided ConfigError.
- Guided errors that recommend `-P effort=none` as an omission spelling (e.g. `_llm.py` /v1/responses hint) reworded to "omit `-P effort=`".
- `AgentPolicyConfig.effort` default becomes `None` and records the resolved value, so the eval log shows the truth.

[...]

PR #335, opened 2026-08-07
fix(capx): effort passthrough parity with the agent plugin (#319)

### LINKED ISSUE #319: capx: effort semantics diverge from agent plugin after #317 (unset injects "low", None means omit)
`plugins/inspect-robots-capx/src/inspect_robots_capx/policy.py:120,156` duplicates the old agent-plugin effort handling: unset defaults to an explicit `"low"`, and `effort=None` means "omit the field" — so `-P effort=none` silently yields the provider default, the exact bug #317 fixes for the agent plugin (PR #318).

Bringing capx to the same contract needs the `_Unset` sentinel (its constructor currently can't distinguish unset from explicit `None`) plus the same normalization (`None` → `"none"`), and a check of which wires capx speaks for the messages-wire `thinking: disabled` mapping.

Deferred from plan 0049 to keep that PR single-plugin.

anishfyi:
## Summary

Brings the capx plugin's `-P effort=` to the contract the agent plugin gained in #318, which plan 0049 deferred and filed as #319.

What the operator types is what the model gets:

| Operator input | Before | After |
|---|---|---|
| flag omitted | `reasoning_effort: "low"` injected | field omitted, provider default applies |
| `-P effort=none` | field omitted, provider default applies | `reasoning_effort: "none"` (chat), `reasoning: {"effort": "none"}` (responses) |
| `-P 'effort="none"'` | `"none"` | unchanged |
| `-P effort=low` .. `max` | passed through | unchanged |
| `-P effort=turbo` | raw `ValueError` traceback | guided `ConfigError` |

Row 2 is the bug worth the release. The core `-P` parser coerces the literal `none` to Python `None` (`defaults.py::_parse_value`), and `CapxPolicy.__init__` read `None` as "omit the field". So an operator asking for **zero** reasoning silently got the provider's **default**, which plan 0049 notes is documented as `high` for Inkling. That is the same inversion #317 describes, and capx kept it because it had no way to tell an omitted flag from an explicit `None`.

Verified against `main` at c8ac6f0 before touching anything:

```
operator input                     wire         eval log
--------------------------------------------------------------
(flag omitted)                     'low'        'low'
-P effort=none  (parses to None)   '<omitted>'  None
-P 'effort="none"' (quoted)        'none'       'none'
-P effort=high                     'high'       'high'

-P effort=turbo -> ValueError: effort must be one of [...], or None to omit the field, got 'turbo'
```

Same script on this branch:

```
operator input                     wire         eval log
--------------------------------------------------------------
(flag omitted)                     '<omitted>'  None
-P effort=none  (parses to None)   'none'       'none'
-P 'effort="none"' (quoted)        'none'       'none'
-P effort=high                     'high'       'high'

-P effort=turbo -> ConfigError: effort must be one of ['high', 'low', 'max', 'medium', 'minimal', 'none', 'xhigh'], got 'turbo'.
fix: omit -P effort= to use the provider default
```

## Changes

- **`policy.py`:** adds the `_Unset` / `_UNSET` sentinel so the constructor can distinguish an omitted `effort` from an explicit `None`. Unset resolves to `None`, which both wire clients already treat as "omit". A supplied value normalizes `None` to the level `"none"` before validation.
- **`CapxPolicyConfig.effort`:** defaults to `None` and records the **normalized** value, so the eval log shows what actually went on the wire rather than what was typed.
- **Error type:** an invalid level now raises `ConfigError` instead of `ValueError`, so the CLI renders a guided message rather than a traceback (the #168 contract, and what the agent plugin does). The old "or None to omit the field" wording was describing behavior that no longer exists, so the message had to change regardless.
- **Tests:** 10 new cases across both wires covering unset, `None`, quoted `"none"`, a named level, the guided error, and a regression test that the sentinel can never leak into a serialized config.
- **Docs:** README argument table, a migration note, and the troubleshooting bullet. CHANGELOG entry under Unreleased.
- **Version:** `inspect-robots-capx` 0.2.0 to 0.3.0, with the matching `uv.lock` line so `uv sync --locked` stays green.

## Two questions from #319, answered

**"Which wires does capx speak for the messages-wire `thinking: disabled` mapping?"**

Only `chat` and `responses`. `_WIRE_FORMATS` is `frozenset({"chat", "responses"})` and the constructor builds only `ChatClient` or `ResponsesClient`, so the Anthropic `thinking: {"type": "disabled"}` mapping #318 added in `_anthropic.py` is unreachable from capx. It is deliberately not ported.

**Do any wire clients need changing?**

No. `ChatClient.complete` (`_llm.py:253`) and `ResponsesClient.complete` (`_responses.py:71`) both already omit the field on `None` and pass the string `"none"` through untouched. The entire fix is in `capx/policy.py` plus docs and tests.

## Breaking change

Scripts that omit `-P effort=` ran at `low` and now run at the provider default. One-line fix: add `-P effort=low` to pin the old behavior. Programmatic callers passing `effort=None` previously meant "omit" and now mean the level `"none"`. This mirrors the break #318 shipped for the agent plugin at 0.23.0, and is called out in both the CHANGELOG and the README.

## Judgement calls, where you could reasonably disagree

1. **`ConfigError` over `ValueError`.** `ConfigError` does not subclass `ValueError`, so this is a small break for anyone catching `ValueError` around construction. I took parity with the agent plugin and the guided-CLI contract as the stronger pull, and the message had to be rewritten anyway. Say the word and I will keep `ValueError`.
2. **Fractional effort is out of scope.** The agent plugin accepts a float in `[0.0, 1.0)` from #315. capx stays named-levels-only because #319 does not ask for it and it widens the diff. Happy to file or send it separately.
3. **The sentinel is duplicated, not shared.** capx only imports the agent plugin's public API today, so reaching for a private `_UNSET` felt worse than a local copy with a sync comment, following the `_PRIOR_LEARNINGS_TEXT_LIMIT` precedent. If you would rather promote the sentinel to shared public API, that is a different and larger change.
4. **`wire` still raises `ValueError`.** Left alone as out of scope, which does leave one function raising two error types. Easy to fold in if you want it.

## Verification

Ran the `plugin-capx` CI job commands locally, plus the neighboring suites to confirm nothing regressed.

```
$ uv sync --locked --all-packages --extra dev
Resolved 100 packages, Checked 86 packages

$ uv run --no-sync ruff check plugins/inspect-robots-capx
All checks passed!

$ uv run --no-sync ruff format --check plugins/inspect-robots-capx
12 files already formatted

$ uv run --no-sync mypy --config-file plugins/inspect-robots-capx/pyproject.toml \
    plugins/inspect-robots-capx/src/inspect_robots_capx plugins/inspect-robots-capx/tests
Success: no issues found in 12 source files

$ uv run --no-sync python -m pytest plugins/inspect-robots-capx/tests -q \
    --cov=inspect_robots_capx --cov-report=term-missing --cov-fail-under=0
79 passed in 7.09s

$ uv run --no-sync python -m pytest plugins/inspect-robots-agent/tests -q --cov-fail-under=0
483 passed in 6.11s

$ uv run --no-sync python -m pytest tests -q
1327 passed, 5 skipped, 1 warning in 18.55s
```

## Checklist

- [x] `pre-commit install` done; hooks pass (ran the `plugin-capx` CI job commands directly, since the root config does not cover plugins)
- [x] Tests added/updated
- [ ] Coverage stays at **100%** (`pytest --cov`) - not applicable, plugin coverage is report-only and the root 100% gate is scoped to `inspect_robots`. The core suite is untouched and still green.
- [x] `mypy` passes (strict)
- [ ] Public API changes reflected in `inspect_robots.__all__` and the API-snapshot test - not applicable, no core public API change
- [x] Core stays NumPy-only (no core files touched)

## Related

Closes #319. Follow-up to #318 (plan [0049](https://github.com/robocurve/inspect-robots/blob/main/plans/0049-effort-passthrough.md)), which fixed the same inversion for the agent plugin and explicitly deferred capx.


anishfyi:
@jeqcho @aris-zhu this is the capx follow-up plan 0049 deferred and you filed as #319, so tagging you for review.

Two things worth your eye specifically:

1. **The wire question from the issue is answered in the PR body.** capx's `_WIRE_FORMATS` is `{"chat", "responses"}` only, so the `thinking: {"type": "disabled"}` mapping #318 added in `_anthropic.py` is unreachable here. I did not port it. If capx is meant to grow a messages wire later, that mapping comes with it rather than now.

2. **I raised the invalid-effort error from `ValueError` to `ConfigError`** so the CLI gives a guided message instead of a traceback, matching the agent plugin and the #168 contract. `ConfigError` does not subclass `ValueError`, so it is a small break for anyone catching `ValueError` around construction. The message needed rewriting anyway (the old "or None to omit the field" wording describes behavior that no longer exists), so it was a one-word extra diff. Happy to revert that half if you would rather keep `ValueError`.

Fractional effort (#315 parity) is deliberately out of scope. Say the word and I will send it as a separate PR.

Ran the `plugin-capx` CI job commands locally plus both neighboring suites: 79 capx, 483 agent, 1327 core, all green, with `uv sync --locked` clean after the 0.3.0 bump.

jeqcho (approved):
Thanks for a really thorough fix — the PR description alone (before/after wire tables, verified against `main` at c8ac6f0) made this a pleasure to review, and the answers to both open questions from #319 check out against the code.

What I verified:

- **The core inversion is real and fixed.** `defaults.py::_parse_value` does coerce a bare `none` to Python `None` (src/inspect_robots/defaults.py:62), and the old constructor read `None` as "omit", so `-P effort=none` silently ran at the provider default. The `_Unset` sentinel in `plugins/inspect-robots-capx/src/inspect_robots_capx/policy.py` resolves this the same way the agent plugin does.
- **True parity with the agent plugin's contract.** The resolution logic (`resolved_effort = "none" if effort is None else effort`, unset → `None` → field omitted) mirrors `inspect_robots_agent/policy.py:414-416` on `main`, and `ConfigError` with a `fix:` line matches the guided-CLI contract from #168. Since capx borrows the agent plugin's `ChatClient`/`ResponsesClient` directly, I confirmed both already omit on `None` and pass `"none"` through (`_llm.py:253`, `_responses.py:71`) — so the whole fix correctly lives in `capx/policy.py`.
- **The messages-wire question from #319.** Confirmed `_WIRE_FORMATS = frozenset({"chat", "responses"})` and only those two clients are constructed, so the Anthropic `thinking: disabled` mapping is indeed unreachable from capx. Good call not porting it.
- **Tests cover the contract, not just the code.** Both wires asserted at the request-body level, the recorded config asserted alongside (so the eval-log normalization claim is pinned), plus the sentinel-leak regression test. The `capx_policy(**kwargs)` factory means an omitted flag genuinely reaches the constructor as unset.
- **Docs/CHANGELOG/version.** Entry sits under `## [Unreleased]` next to the agent 0.23.0 entry it mirrors; README table and troubleshooting bullet updated; 0.3.0 bump with matching `uv.lock`. CI is fully green including `plugin · capx`.

On the judgement calls: I agree with all four. `ConfigError` is the right pull given the message needed rewriting anyway; fractional effort and the `wire` `ValueError` are cleanly out of scope for #319 (worth a small follow-up issue for the latter if you feel like it — one constructor raising two error types is mildly untidy but harmless); and the duplicated sentinel with a sync comment beats importing a private symbol.

Nice work — this is exactly the shape of PR that's easy to trust.

jeqcho (approved):

PR #349, opened 2026-08-08
fix(voice): playback-aware muting when --speak and --voice run together (#332)

### LINKED ISSUE #332: voice: playback-aware muting when --speak and --voice run together
Follow-up from #327 / PR #328 (plan 0054, design decision 10).

With `--speak` and `--voice` both enabled, the microphone can pick up the TTS narration and feed it back into the operator-message channel: the voice input's energy gate and hallucination guards are tuned to accept speech, so they will not filter another speaker's speech. Docs currently tell the operator to separate mic and speaker or use a headset.

Proper fix: while the SpeakerSink has an utterance in flight, the voice input should discard (or gate) captured segments that overlap playback. Needs a small shared signal between the two components inside the plugin (both live in inspect-robots-voice, so no core seam is required), plus a test that a segment timestamped inside a playback window is dropped.

Adityakk9031:
Closes #332

### Description

With both `--speak` and `--voice` enabled, the microphone can capture the TTS playback and feed it back into the operator-message channel.

This PR implements a lightweight, thread-safe synchronization mechanism inside the `inspect-robots-voice` plugin to drop microphone audio blocks captured during active TTS playback and during an echo-tail hangover period.

### Changes

- **`inspect-robots-voice` (version 0.5.2)**:
  - **Thread-safe registry & hangover timer**: Added `_active_speakers` set, `_last_playback_end` timestamp, and `_HANGOVER_SECONDS = 0.3` guarded by `_speakers_lock` in `_capture.py`.
  - **Drop muted blocks**: Updated `MicrophoneCapture._callback` to discard blocks immediately when playback or hangover is active (`_is_playback_active()`), freezing the energy gate state and preventing noise floor collapse.
  - **Clean speaker integration**: Hoisted `_register_speaker` and `_unregister_speaker` imports to the top of `_speaker.py` and wrapped playback writes in `try...finally` to guarantee unregistration and hangover timestamp recording.
  - **Path escaping**: Handled path string regex escaping (`re.escape`) in `test_tts.py` for Windows paths.
  - **Version bump**: Bumped plugin version to `0.5.2` with clean `uv.lock` containing zero environment-marker churn.

### Verification & Testing

- Added `test_playback_aware_muting_callback` in `tests/test_capture.py` verifying that blocks are dropped during playback and during echo-tail hangover, and preserved normally afterward.
- Added `test_is_playback_active_with_explicit_timestamps` verifying the hangover duration logic.
- Added `test_speaker_active_playback_registration` in `tests/test_speaker.py` verifying registration bracketing around chunked playback writes.
- Verified all 199 voice plugin tests pass and ruff lint/format checks pass.
- Added CHANGELOG entry under `### Fixed`.


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks @Adityakk9031 for taking on #332 — the mechanism you built is the right shape (a small shared signal between `SpeakerSink` and `MicrophoneCapture`, exactly what the issue asked for), the `try/finally` bracketing is careful, and the tests for registration and muting are clear. Two behavioral issues need addressing before this fixes the bug end to end, though:

**1. Zero-filling collapses the adaptive noise floor (blocking)**

The muted blocks still flow into `EnergyGate.push`, and a zero block has RMS 0. While the gate is closed, `_push_closed` in `plugins/inspect-robots-voice/src/inspect_robots_voice/_segmenter.py` runs

```python
self._noise_floor = (1.0 - self.ema_alpha) * self._noise_floor + self.ema_alpha * rms
```

so each 0.1s zero block multiplies the noise floor by 0.95. A 10-second narration is ~100 blocks, shrinking the floor ~170x. When the mute lifts, ordinary room noise is far above the near-zero threshold, so the gate opens immediately — and since the floor only re-adapts while the gate is *closed*, it can stay open until the 30s `max_utterance_s` cap and ship a long noise "utterance" to the transcriber. That's the very hallucination path this PR is meant to close, so the description's claim that zeroing "keeps the adaptive energy gate statistics stable" is unfortunately inverted. Options, roughly in order of preference:

- **Drop** muted blocks in `MicrophoneCapture._callback` (`_capture.py` ~line 138) instead of enqueueing zeros — this matches the issue's "discard captured segments" wording and freezes the gate state during playback; or
- keep zero-filling but have the segmenter skip noise adaptation for muted blocks; or
- `reset()` the gate when playback ends.

**2. Echo tail leaks after unmute (blocking, smaller)**

`_active_speakers.discard(self)` in `_speaker.py` fires as soon as the last blocking `playback.write` returns, but output-stream latency plus room acoustics plus the 0.1s input blocksize mean the tail of the narration is still arriving at the mic for a few hundred milliseconds afterward — and there are brief unmuted gaps between queued utterances (register/discard is per queue item). A short mute hangover would cover both: e.g. record `last_playback_end = time.monotonic()` on discard and treat capture as muted while a speaker is active *or* within ~0.3–0.5s of the last one. That would also let you add the test the issue asks for (a segment overlapping a playback window is dropped, including its tail).

**Smaller points**

- `_speaker.py` (~line 312): the `from inspect_robots_voice._capture import ...` sits inside the worker loop and re-executes every iteration. There's no import cycle (`_capture` doesn't import `_speaker`), so please hoist it to the module top.
- `uv.lock`: ~100 lines of environment-marker churn unrelated to the version bump — looks like it was regenerated with a different `uv` version. Could you re-lock so only the `inspect-robots-voice` version entry changes?
- `tests/test_tts.py`: the `re.escape` fix is sensible (Windows paths, I'd guess) but unrelated to this PR — worth a mention in the description or splitting out so the changelog story stays clean.
- Taking a `threading.Lock` inside the PortAudio callback is fine here given both critical sections are tiny, so no change needed — just noting it was considered.

Really appreciate the thorough write-up and the passing CI matrix — with the drop-instead-of-zero change and a small unmute hangover this will land nicely.

Adityakk9031:
Hi @jeqcho, thanks for the insightful feedback on the adaptive noise floor dynamics and echo tail latency! I have rebased onto `main` and updated the implementation:

1. **Drop instead of zero-fill**: Updated `MicrophoneCapture._callback` to return early without enqueueing when playback is active. This freezes the gate state without decaying the adaptive noise floor to zero.
2. **Echo tail hangover**: Added `_last_playback_end` timestamp tracking and `_HANGOVER_SECONDS = 0.3` to keep capture muted while a speaker is active *or* within 0.3s of playback completion.
3. **Module-top imports**: Hoisted `_register_speaker` and `_unregister_speaker` from `_capture.py` to the top of `_speaker.py`.
4. **Clean `uv.lock`**: Restored `uv.lock` from `main` and bumped only `inspect-robots-voice` to `0.5.2` without any environment-marker churn.
5. **Tests**: Updated `test_playback_aware_muting_callback` to assert block dropping (queue stays empty) during active playback and during hangover, plus `test_is_playback_active_with_explicit_timestamps` for timing assertions. Noted the `re.escape` in `test_tts.py` in the PR description.

All 199 voice plugin tests pass and ruff lint/format checks pass. Ready for re-review!

Adityakk9031:

1. **Playback buffer latency & drain:** Implemented `drain()` on `_SoundDevicePlayback` and added buffered playback drain/wait logic in `_worker` before unregistering the speaker. This prevents microphone un-muting while audio remains in device hardware buffers (up to 500ms), ensuring echo tail hangover is held until playback actually completes.
2. **Tests & Changelog:** Added regression test `test_speaker_buffered_playback_holds_mute_until_drained` in `plugins/inspect-robots-voice/tests/test_speaker.py` and resolved `CHANGELOG.md` merge conflict. All 200 tests and all 20 CI jobs are green (`ci-ok pass`).

PR #373, opened 2026-08-12
fix(agent): sanitize scene IDs in wire capture and transcript paths (#370)

### LINKED ISSUE #370: Agent plugin side-cars use raw scene ids in filesystem paths (transcripts + wire capture)
Found during plan 0067 review (see plans/0067-durable-action-log.md, Design → File layout).

`LLMAgentPolicy` builds side-car paths from the raw scene id in two places:

- `on_trial_start` passes `f"{scene_id}-e{epoch}"` raw to `_capture.begin_trial` (wire capture path)
- `on_trial_end` writes `transcripts/<run_id>/{scene_id}-e{epoch}.jsonl` with the raw id

`Scene.id` is unvalidated task-author input, so a hostile or merely unusual id (path separators, `..`) lands directly in a filesystem path. Core's `FrameStore` already solves this with `_safe` (unsafe-run substitution + crc32 suffix so distinct ids can't collide), and plan 0067's `actions/` side-car reuses it — the plugin's two paths should too.

Low urgency: every in-tree task uses tame ids today. But it's two call sites and the helper already exists.

Adityakk9031:
Fixes: #370 
### Description

`LLMAgentPolicy` was constructing transcript and wire capture filesystem paths from the raw `scene_id` value. Since `scene_id` is unvalidated task-author input, unusual strings or directory traversal patterns could write files outside the intended log directories.

This PR sanitizes the scene IDs using the plugin's `_safe` helper (matching core's `FrameStore` and `eval.py` sanitization logic) before constructing side-car paths and metadata pointers.

### Changes

- **`inspect-robots-agent`**:
  - Imported `_safe` from `_capture` in `policy.py`.
  - Modified `on_trial_start` to pass `_safe(scene_id)` instead of `scene_id` to `_capture.begin_trial`.
  - Modified `on_trial_end` to construct the `trial_id` transcript filename with `_safe(record.scene_id)`.
  - Added a new test `test_scene_id_sanitization_prevents_directory_traversal` in `tests/test_policy_e2e.py` to assert that paths are safely resolved within the target run directory and directory traversal is blocked.
  - Documented the fix in `CHANGELOG.md`.

### Verification

- Ran the agent plugin test suite:
  - `pytest plugins/inspect-robots-agent/tests` -> 486 passed.
- Formatting and check verification:
  - `ruff check plugins/inspect-robots-agent` -> Passed.
  - `ruff format --check plugins/inspect-robots-agent` -> Passed.


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks @Adityakk9031 — this is a tidy take on #370. You picked up exactly the helper the issue pointed at, kept the diff to the two call sites, and the transcript stem you produce (`f"{_safe(record.scene_id)}-e{record.epoch}"`) is character-for-character what core's `eval.py:93` uses for the `actions/` side-car, so for an unusual scene id the `actions/`, `transcripts/`, and (now) `wire/` entries all share one stem. I also traced every consumer (`cli.py:_load_wire_trials`, `_html.py:1094`) and they all follow the `wire_capture` / `transcript` metadata pointers via `resolve_log_pointer` rather than recomposing names, so the rename is safe for existing tooling. Suite is green locally: 486 passed, ruff and mypy clean.

Two things before merge, then a couple of nits you can fold into the same push.

**Blocking**

1. **Rebase onto `main`** — the PR is 11 commits behind and conflicts on `CHANGELOG.md` (policy.py merges cleanly). While you're there, please match the existing entry style under `## [Unreleased] → ### Fixed`: prefix with the plugin version like the neighbouring entry (`**Agent plugin (0.26.0):**`), and wrap at ~80 columns like the rest of the file.

2. **Make the CHANGELOG / commit message accurate about what changed.** The entry says the fix prevents directory traversal in *wire capture and transcript paths*, but the wire path was never traversable: `_capture.py:93` already runs `_safe(trial_id)` inside `begin_trial`. What your `on_trial_start` change actually does is move the sanitisation before the `-e{epoch}` suffix, so the wire directory becomes `unsafe-..-id-<crc>-e0` (matching transcripts and core) instead of `unsafe-..-id-e0-<crc>`. That is a real and welcome improvement, but the user-facing note should say "transcript paths are now sanitised; wire-capture directories are renamed to share the same stem as `transcripts/` and `actions/`" rather than implying both were traversal bugs. Users reading the changelog to decide whether to rotate logs deserve the precise version.

**Nits (non-blocking)**

3. `plugins/inspect-robots-agent/tests/test_policy_e2e.py:2798` — `unsafe_id = "unsafe/../id"` does not actually escape the run directory even when left raw: `transcripts/run-1/unsafe/../id-e0.jsonl` resolves *inside* `run-1`, so the `is_relative_to(...)` assertions would pass on the unsanitised path too. The test only fails pre-fix because `unsafe/` doesn't exist and `open()` raises `FileNotFoundError` (I verified this by running it against the merge-base `policy.py`). An id like `"../escaped"` makes the `is_relative_to` checks do the real work and documents the threat more honestly.

4. Same test, the wire-capture half (`:2811–2815`) passes without your change for the reason in (2). If you want it to guard the behaviour you actually introduced, assert that the wire directory name and the transcript stem agree, e.g. `Path(wire_ptr).parts[2] == Path(transcript_ptr).stem`. That would fail on pre-fix code and pins the alignment that is the point of the `on_trial_start` change.

5. The 5 pre-existing mypy errors in `test_policy_e2e.py` (lines 1998–2529) are not yours and CI doesn't type-check plugin tests, so no action needed — just flagging so you don't chase them.

Once the rebase and the wording are in I'm happy to merge.

Adityakk9031:

1. Rebased onto latest `main` (resolving conflict on `CHANGELOG.md`).
2. Updated `CHANGELOG.md` entry and commit message to accurately describe what changed: transcript paths are now sanitised; wire-capture directories are renamed to share the same stem as `transcripts/` and `actions/`.
3. Updated test to use `unsafe_id = "../escaped"` so `is_relative_to` genuinely exercises directory containment.
4. Added assertion that the wire capture directory name and transcript stem agree (`Path(wire_ptr).parts[2] == Path(transcript_ptr).stem`).

Adityakk9031:

1. **Rebase & Changelog:** Rebased onto `main`, resolved `CHANGELOG.md` conflict with accurate entry style under `## [Unreleased] -> ### Fixed` (`**Agent plugin (0.26.0):**`).
2. **Accurate description:** Updated changelog and commit message to specify that transcript paths are sanitized and wire-capture directories are renamed to share the same stem as `transcripts/` and `actions/`.
3. **Escaped traversal test:** Updated test to use `unsafe_id = "../escaped"` to verify path escape rejection.
4. **Path alignment:** Added assertion that wire directory name matches the transcript stem (`Path(wire_ptr).parts[2] == Path(transcript_ptr).stem`). All 20 CI checks are passing (`ci-ok pass`).

jeqcho (approved):

PR #374, opened 2026-08-12
fix(controller): handle empty action chunks gracefully by raising PolicyError

Adityakk9031:
Summary: Fixes unhandled IndexError and ValueError when a policy returns an empty ActionChunk. Raises PolicyError instead.

### DIFF
--- a/src/inspect_robots/controller.py
+++ b/src/inspect_robots/controller.py
@@ -23,6 +23,7 @@ from typing import Any, Protocol, runtime_checkable
 
 import numpy as np
 
+from inspect_robots.errors import PolicyError
 from inspect_robots.policy import Policy
 from inspect_robots.spaces import Box
 from inspect_robots.types import Action, Observation
@@ -58,6 +59,8 @@ class DefaultController:
         buffer: deque[Action] = store.setdefault(_BUFFER_KEY, deque())
         if not buffer:
             chunk = policy.act(observation)
+            if not chunk.actions:
+                raise PolicyError("policy emitted an empty ActionChunk (0 actions)")
             take = self.replan_interval or len(chunk)
             taken = list(chunk.actions)[:take]
             buffer.extend(taken)
@@ -163,6 +166,8 @@ class EnsemblingController:
     ) -> Action:
         """Query once for step ``t`` and blend every retained prediction for that step."""
         chunk = policy.act(observation)
+        if not chunk.actions:
+            raise PolicyError("policy emitted an empty ActionChunk (0 actions)")
         store.setdefault(_INFER_KEY, []).append((chunk.inference_latency_s, len(chunk)))
 
         buffer: list[tuple[int, list[Any], dict[str, Any]]] = store.setdefault(_ENSEMBLE_KEY, [])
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks for the PR, @Adityakk9031. The instinct is right: a controller should never hand the rollout an `IndexError`/`ValueError` when the policy misbehaves, and it should surface as a `PolicyError` so the trial is recorded as failed rather than crashing the eval. You also kept the change small, added a test per controller, and the branch is green on every gate (ruff, mypy strict, 100% coverage). Appreciated.

The problem is that both halves of this are already guaranteed by code that exists today, so the diff adds guards that cannot fire. Walking through it:

**Blocking**

1. `src/inspect_robots/controller.py:62-63` and `:169-170`: a policy cannot return an empty chunk. `ActionChunk.__post_init__` at `src/inspect_robots/types.py:79-81` raises `ValueError("ActionChunk must contain at least one action")` on construction, and that invariant is already pinned by `tests/test_types_spaces.py:62`. The new test has to reach for `object.__setattr__(chunk, "actions", [])` on a frozen dataclass to fabricate the state, which is the tell: the state it exercises is not reachable through the public API. A test that has to break a frozen dataclass's invariant to trigger the branch is testing a situation the framework already forbids.

2. Even the fallback case is already covered. If a policy tries to build an empty chunk, the `ValueError` propagates out of `policy.act()` into `controller.next_action`, and `src/inspect_robots/rollout.py:363-368` catches every non-`InspectRobotsError` exception there and re-raises it as `PolicyError` via `_policy_error(policy, exc)`, recording the failure on the trial. That is the documented contract in `src/inspect_robots/CLAUDE.md` ("Generic policy/embodiment exceptions ... are wrapped into `PolicyError`"). So the `IndexError`/`ValueError` the PR description says are unhandled never escape the rollout today.

3. `src/inspect_robots/controller.py:187-188`: the `if not predictions:  # pragma: no cover` guard is dead code. The chunk for the current step was appended two lines earlier with `q == t`, and after the empty check `len(acts) >= 1`, so `0 <= t - q < len(acts)` always retains it and `predictions` is never empty. Adding `pragma: no cover` to a newly written branch to get it past the 100% gate is the opposite of how this repo uses that marker; the one existing pragma in this file (`:148`) sits on a branch that a comment proves unreachable and that pre-dates the check. Please do not add code whose only way past the coverage gate is an exclusion.

Given 1-3, my ask is to close this PR rather than rework it. If you have actually seen a raw `IndexError` or `ValueError` escape a rollout from an empty chunk (a traceback, a plugin, a specific policy), please open an issue with the repro and I will happily look, because that would mean the wrapping at `rollout.py:363-368` has a hole and the fix belongs there, not in the controllers.

**Nits** (moot if closed, noted for future PRs)

- `tests/test_controller.py:124-127`, `:139-143`: `import pytest` and `from inspect_robots.errors import PolicyError` belong at module top with the other imports, not inside each test function.
- No `CHANGELOG.md` entry under `## [Unreleased]`; user-visible behaviour changes need one here.

Adityakk9031:
Closing as requested - thank you for the review and explanation!

PR #375, opened 2026-08-12
fix(frames): cap safe filename lengths in FrameStore and json_log to prevent OS path overflow

Adityakk9031:
Summary: Prevents FileNotFoundError/OSError on long scene IDs and camera names by capping slug length in FrameStore and json_log.


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks @Adityakk9031 — this closes a genuine hole. `FrameStore.put` had no cap at all, so a long `scene.id` or camera name would blow past the 255-byte filename limit inside `np.save` mid-rollout, and the new test (`tests/test_rollout_hardening.py:797`) does fail without the fix (a 300+300-char name yields a 612-byte filename). Keeping the CRC32 of the *original* name as the suffix is the right call: I checked that two 61-char names differing only in the last character still land on distinct files. Ruff, mypy, and the full suite are all clean.

The problem is that `_safe` isn't only used to *write* filenames — `_html.py` re-derives them to *find* frames on disk, so any change to its output for names that previously worked breaks report rendering for already-recorded logs. A few things need to change before this can land.

**Blocking**

1. **`src/inspect_robots/frames.py:22,32-34` — the cap renames files that were never at risk.** `_SAFE_MAX = 50` changes the output for every name over 50 characters, but a 60-char scene ID produced an 80-byte filename under the old code, well within limits. `_html.py:677`, `:1023`, `:1070`, and `:1537` recompute `_safe(f"{scene_id}-e{trial}")` and `_safe(camera)` to locate `.npy` files, so a report rendered with this change for a log recorded before it (scene ID ≥ ~47 chars) will silently find no frames — `_frame_path` returns `None` on `path.exists()` and the flipbook goes empty with no error. Please raise the cap so it only bites where the old code would already have failed. Two components at 100 each give a worst-case `100+9+1+100+9+1+6+4 = 230` bytes, still under 255, and leave every name in the 51–100 range byte-identical to today. Add a `CHANGELOG.md` line under "Unreleased → Fixed" noting that names above the new cap now change on disk, since that is a visible behaviour change for anyone with long scene IDs.

2. **`src/inspect_robots/frames.py:34` — `rstrip("-")` changes short names too.** It runs on the "anything replaced" branch as well as the truncated one, so a name that merely ends in an unsafe character flips: `_safe("scene/")` was `scene--1f0c90f1` and is now `scene-1f0c90f1`; `_safe("top camera ")` goes from `top-camera--081496b9` to `top-camera-081496b9`. Same lookup breakage as above, plus `eval.py:93` uses `_safe(record.scene_id)` for the action-log pointer filename. Either drop the `rstrip` or apply it only when `len(safe) > _SAFE_MAX` actually truncated something (and even then it is cosmetic — a double dash before the hash is harmless).

3. **`src/inspect_robots/logging/json_log.py:32-34` — please drop the `_SLUG_MAX` 200 → 100 hunk.** The eval-log filename was already capped and already fits: `tests/test_strict_json.py:116` (`test_long_task_name_still_writes_its_log`, #292) runs a 300-char task name against the 200 cap and asserts the name stays ≤ 255 bytes. The new value is not exercised by any test, and the rewritten comment ("headroom for Windows MAX_PATH limits") is not accurate — `MAX_PATH` is a 260-char limit on the *full path*, so shortening one basename by 100 characters gives no guarantee, and the Windows job in `test-extra` was already passing on main. If there is a concrete failing case behind this, a test that reproduces it would justify the change; otherwise it is unrelated to the FrameStore bug this PR fixes.

**Tests (blocking, goes with the above)**

4. **`tests/test_rollout_hardening.py:797`** — the new test proves the file is writable but not the two properties the docstring promises. Please add: (a) a collision check — two long names identical for the first `_SAFE_MAX` chars map to different paths; and (b) a regression guard that short names are unchanged, e.g. `assert _safe("a/b") == "a-b-" + <crc>` and `assert _safe("x" * 50) == "x" * 50`, so a future edit cannot quietly shift on-disk names again. `_safe` is already imported directly in `tests/test_html_view.py:35` and `tests/test_eval_action_log.py:17`, so importing it here matches existing practice.

**Nits (non-blocking)**

- The new test lives at the very end of the file after the `PolicyError` tests; the other `FrameStore` tests sit around `:253-262`. Moving it next to them keeps the file's sections coherent.
- `CONTRIBUTING.md:131` and the PR template ask for a `CHANGELOG.md` entry under "Unreleased" — tick that box once item 1 is in.

Happy to re-review quickly once the cap and the `rstrip` are sorted; the core idea here is sound.

Adityakk9031:

1. Raised `_SAFE_MAX = 100` in `src/inspect_robots/frames.py`, so filenames under 100 characters remain byte-identical to existing logs on disk.
2. Avoided `rstrip("-")` on short names so suffixes like `scene/` and `top camera ` produce identical hashes and filenames as before.
3. Dropped the `_SLUG_MAX` hunk in `json_log.py`.
4. Moved the tests next to other `FrameStore` tests in `tests/test_rollout_hardening.py`, and added collision checks for names identical on the first 100 characters, regression checks for short names and trailing unsafe characters, and length bounds.
5. Added `CHANGELOG.md` entry under `[Unreleased] -> Fixed`.

Adityakk9031:

1. **Safe filename length:** Kept `_SAFE_MAX = 100` for filesystem portability across ext4, NTFS, and APFS.
2. **Slug characters preserved:** Removed `rstrip("-")` to avoid mutating valid slug ends.
3. **Scoped diff:** Dropped `_SLUG_MAX` hunk from `frames.py`.
4. **Tests & CI:** Expanded test in `tests/test_rollout_hardening.py` covering collision-free truncation and fixed isort separation. 100% test coverage gate and all CI checks are passing (`ci-ok pass`).

PR #376, opened 2026-08-12
fix(approver): reject non-finite values in DeltaLimitApprover to prevent reference state corruption

Adityakk9031:
Summary: Prevents infinite value propagation and reference state corruption in DeltaLimitApprover by raising SafetyAbort when non-finite values are encountered.


Adityakk9031:
@jeqcho have a look

jeqcho (changes requested):
Thanks @Adityakk9031, this is a real gap and I'm glad you caught it. I confirmed it against `main`: with `DeltaLimitApprover(_abs_space(), max_delta=0.1)` and an empty store, reviewing `[inf, 0.0]` passes through as the first action, gets stored as the reference, and the very next action `[0.5, 0.0]` comes back as `[inf, 0.0]` because `np.clip(0.5, inf - 0.1, inf + 0.1)` is `inf`. From then on the limiter is silently dead on that dimension. In the CLI wiring `ClampApprover` normally runs first and absorbs the `inf` on bounded boxes, but a box that is unbounded on one side (which the delta limiter explicitly accepts when `max_delta` is given) or a standalone `DeltaLimitApprover` walks straight into this. Raising `SafetyAbort` is the correct response, the test covers both signs, and the change is tight and readable. Nice work.

A few things before I can merge, split into blocking and nits.

**Blocking**

1. `src/inspect_robots/approver.py:255` — the `rewind_reference` change silently swallows a non-finite pose. That inverts the safety logic of the rest of the PR: if a downstream approver hands us a pose containing `inf`/`NaN` as "the pose that actually ran", something upstream is already badly broken, and quietly leaving the old reference in place means the next delta is measured from a pose the robot is not at. Please either drop this hunk (the `review()` guard already prevents a non-finite value from ever entering the store through the normal path) or make it raise (`ValueError` is the convention for caller bugs in this file, `SafetyAbort` if you want it treated like a poisoned action). Whichever you pick, add a test; right now the new branch has no coverage of the non-finite path at all.

2. `src/inspect_robots/approver.py:162` — the class docstring still says only `NaN` raises `SafetyAbort`. That docstring is the generated API reference, so it needs to say `±inf` aborts too. Please also add one sentence on *why* this differs from `ClampApprover`, whose docstring at `approver.py:83-88` deliberately documents "`±inf` is *not* an abort: it clamps to the finite bound". The reason is exactly the one you found: a delta limiter has no trustworthy finite bound on the first absolute step, so it cannot clamp `inf` into anything meaningful. Without that sentence the two approvers look inconsistent by accident rather than by design.

3. `CHANGELOG.md` — add an entry under `## [Unreleased]` / `### Fixed`, `**Core:**` prefix, matching the existing style. Something like: "`DeltaLimitApprover` now raises `SafetyAbort` on `±inf` as well as `NaN`; previously an infinite first absolute-mode target was stored as the reference and disabled the limiter on that dimension for the rest of the trial."

**Nits**

4. `src/inspect_robots/approver.py:263` — "NaN or non-finite values" is redundant since NaN is non-finite. "action contains a non-finite value (NaN or ±inf)" keeps the word `NaN` so `tests/test_approvers.py:411` (`match="NaN"`) still passes, and your new `match="non-finite"` still matches.

5. `tests/test_approvers.py:211` — consider parametrising over `_abs_space()` and `_delta_space()` the way `test_nan_raises_safety_abort_in_both_branches` does at `:408`. Displacement mode is where the behaviour actually changes user-visibly (an `inf` there used to clamp to `±max_delta`, now it aborts), so it deserves an explicit assertion.

6. The commit message trailer "…ent reference state corruption" looks like a wrapped subject line got split into the body; a plain one-line subject is fine.

Once 1-3 are in I'm happy to merge this.

Adityakk9031:

1. Updated `rewind_reference` to raise `ValueError` on non-finite poses rather than silently swallowing them, with dedicated test coverage.
2. Updated `DeltaLimitApprover` class docstring to explicitly document that non-finite values (`NaN` or `?inf`) abort, contrasting with `ClampApprover` and explaining why delta limiters cannot clamp `?inf` meaningfully on the initial step.
3. Added `CHANGELOG.md` entry under `[Unreleased] -> Fixed`.
4. Kept "action contains a non-finite value (NaN or ?inf)" to retain matching `NaN` and `non-finite`.
5. Parametrised `test_inf_raises_safety_abort_in_both_branches` over both `_abs_space()` and `_delta_space()`.
6. Rebased onto `main` and cleaned up commit message.

Adityakk9031:

1. **Raise on non-finite reference:** Updated `rewind_reference` to raise `ValueError` matching the file's caller error convention when given a non-finite pose, and added coverage test.
2. **Docstring:** Updated class docstring to explicitly mention non-finite values (`inf` and `NaN`) and contrast with `ClampApprover`.
3. **Changelog:** Added entry in `CHANGELOG.md` under `## [Unreleased] -> ### Fixed`.
4. **Parametrized tests:** Parametrized tests over `[-np.inf, np.nan]` with `match="non-finite"`. 100% test coverage gate and all CI checks are passing (`ci-ok pass`).

PR #377, opened 2026-08-13
Close registry-owned policies after evaluation

PenTest-duck:
Closes only policies resolved by eval, mirrors existing embodiment ownership, preserves caller-owned policy lifecycles, and guarantees both cleanup paths are attempted on failure. Cleanup remains embodiment-first. The Dropbear adapter does not depend on this contribution.

### DIFF
--- a/src/inspect_robots/eval.py
+++ b/src/inspect_robots/eval.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 import os
 import subprocess
+import sys
 import time
 import uuid
 import warnings
@@ -249,10 +250,10 @@ def eval(
 
     ``task``/``policy``/``embodiment`` may be objects or **registry names**
     (e.g. ``policy="scripted"``), resolved through the registry — the Inspect-style
-    ergonomic that keeps logs and the CLI reproducible. An embodiment resolved
-    from a registry name is owned by ``eval()`` and is closed when the run
-    finishes (even on a halt); a caller-constructed embodiment object stays
-    open — the caller owns its lifecycle.
+    ergonomic that keeps logs and the CLI reproducible. Policies and embodiments
+    resolved from registry names are owned by ``eval()`` and closed when the run
+    finishes (even on a halt); caller-constructed objects stay open — the caller
+    owns their lifecycle.
 
     ``seed=None`` draws a fresh seed from the OS and records it in the log, so
     an "unseeded" run remains reproducible after the fact (and is distinct from
@@ -310,6 +311,7 @@ def eval(
     from inspect_robots.registry import resolve
 
     before_scoring = _grading_hook(grader, before_scoring)
+    owns_policy = isinstance(policy, str)
     owns_embodiment = isinstance(embodiment, str)
     task = cast(Task, resolve("task", task)) if isinstance(task, str) else task
     policy = cast(Policy, resolve("policy", policy)) if isinstance(policy, str) else policy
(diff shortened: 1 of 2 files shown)


jeqcho (changes requested):
Thanks @PenTest-duck, this is a careful piece of work. The core of it is exactly right: `eval()` promised "close what we open" for embodiments and quietly never honoured it for policies, even though `agent`, `xpolicylab`, and `capx` all carry connection pools or sockets behind `close()`. I especially like that you moved resolution inside the `try` so a failing embodiment factory still closes the already-resolved policy, that cleanup is embodiment-first with both closes always attempted, and that a cleanup failure never masks the run's real exception (tracked explicitly via `active_error` rather than `sys.exc_info()`, which the `inside_outer_except` test nails down). I reverted `src/inspect_robots/eval.py` to the merge base and 7 of your 8 new tests fail, so they genuinely pin the behaviour. ruff, mypy, and the 100% coverage gate are all green under CI's install.

The code is approve-quality. What is missing is the paperwork the repo treats as part of the change, since this alters a documented public contract.

**Blocking**

1. **CHANGELOG entry.** Add a line under `## [Unreleased]` in `CHANGELOG.md` (an `### Added` or `### Changed` bullet under **Core:**) saying registry-resolved policies are now closed by `eval()`, mirroring embodiments. Every behaviour change in this repo lands with one; the existing embodiment-ownership entry near `CHANGELOG.md:802` is a good template.

2. **User docs still describe embodiment-only ownership.** You updated the `eval()` docstring, but the guides now contradict it:
   - `docs/guide/concepts.md:95-98` ("an embodiment resolved from a registry name is closed ... a caller-constructed embodiment object stays open")
   - `docs/guide/policies-and-embodiments.md:100-103` (the "Lifecycle" paragraph)
   - `src/inspect_robots/CLAUDE.md:58-59` (the agent-facing invariant list)

   Please widen all three to "policies and embodiments". While you are in `policies-and-embodiments.md`, add one sentence to the policy-authoring section stating that `close()` is an optional hook on a `Policy` (the Protocol at `src/inspect_robots/policy.py:55` does not declare it) and that `eval()` calls it on registry-resolved policies. Right now nothing tells a plugin author this hook exists, which is the whole reason the leak went unnoticed.

**Non-blocking**

3. `src/inspect_robots/eval.py:350-359`: the cleanup catches are `except BaseException`. With an active error already set, a second Ctrl-C arriving during `owned_embodiment.close()` gets downgraded to a note and execution continues into `close_policy()`, which for `xpolicylab` does a network `trial_end` goodbye and could block. Before this PR that interrupt propagated. Consider `except Exception` on both cleanup catches so `KeyboardInterrupt`/`SystemExit` keep their teeth; the safety-relevant close (the embodiment) has already been attempted by that point either way.

4. `src/inspect_robots/eval.py:313-314`: `owns_policy` / `owns_embodiment` are now redundant with `owned_policy is not None` / `owned_embodiment is not None`, since the `owned_*` slots are only assigned on the string path. Dropping the booleans removes two dead conjuncts from the `finally`.

5. `src/inspect_robots/eval.py:365-368`: on Python 3.10 (which pyproject still supports) there is no `add_note`, so a cleanup failure under an active error is silently dropped. `warnings` is already imported in this module; a `warnings.warn(...)` fallback in the `else` branch would keep it visible.

6. The PR body's "The Dropbear adapter does not depend on this contribution" reads as context from elsewhere; happy to drop it from the description.

One thing worth a follow-up issue rather than this PR: the CLI's `_resolve_components` (`src/inspect_robots/cli.py:1429`) constructs the policy itself and hands objects to `eval()`, and its `finally` chain at `cli.py:1736-1745` closes speaker, voice input, and embodiment but never the policy. So `inspect-robots run --policy agent` still leaks the httpx pool after this lands. Not your scope here, but it is the same bug one layer up and now has an obvious shape.

PR #404, opened 2026-08-22
fix: route hook-raised halt-class errors through halted path (#401)

### LINKED ISSUE #401: A SafetyAbort/EmbodimentFault raised outside the rollout discards the run's log
Follow-up from PR #400 review. When observe_parked() (or a before_scoring hook) raises SafetyAbort/EmbodimentFault, the error propagates out of _run_eval before bus.on_eval_end runs, and JsonLogSink writes only at on_eval_end, so the whole run's JSON log is discarded (an e-stop during the pre-grade park on trial 40/40 loses all 40 trials). Rollout-raised halts already route through the halted machinery and preserve the log; the hook paths are the only ones that lose it. Existing precedent: a raising before_scoring behaves the same way, so this is spec-conformant today. Proposal: route hook-raised halt-class errors through the same halted=True / status=error path the rollout uses, then re-raise or record per the documented halt semantics.

Amresh-01:
Closes #401 

### Description
This PR addresses an issue where `SafetyAbort` or `EmbodimentFault` exceptions raised by a `before_scoring` hook (or `observe_parked()`) would completely propagate out of `_run_eval` before `bus.on_eval_end` could run, discarding the JSON log for the entire evaluation run.

### Changes Made
- Wrapped the scoring hook phase (the `before_scoring` call and `scorers` loop) in a `try...except (EmbodimentFault, SafetyAbort)` block.
- Buffered the individual scores into an `epoch_scores` dictionary during iteration to prevent partial scoring data from polluting `per_scorer_scores` if a mid-loop scorer crashes.
- Mirrored the rollout-raised exception path: on catching the exception, we now set `halted = True`, set the run and scene statuses to `"error"`, mark the trial as an error, and record empty/null scores so the trial does not masquerade as valid data.
- The evaluation loop now degrades gracefully, ensuring `bus.on_eval_end` successfully writes the JSON log of all completed and errored trials up to that point.


Amresh-01:
Hi @jeqcho - PR ready for review.

jeqcho (changes requested):
Thanks @Amresh-01, this is a good fix for #401 and the core of it is right. You put the new `except (EmbodimentFault, SafetyAbort)` at the same level as the rollout one and set exactly the same five things (`status`, `error`, `scene_status`, `scene_error`, `halted`), so the existing `if halted: stopped = True; break` at `src/inspect_robots/eval.py:689` does the rest: no further trial starts, the reducers run, `bus.on_eval_end` writes the log, and the `finally` in `eval()` still closes an owned embodiment. I traced the whole tail of the loop and nothing touches the embodiment after the catch, which is the property that matters on a real arm. Marking the record `status="error"` so it lands in the "recorded but never scored" bucket is also the right call, and keeping the inner `except (SafetyAbort, EmbodimentFault): raise` so the generic `observe_parked` fallback cannot swallow a halt was well spotted. All four gates pass locally at 100% coverage.

Four things before I can merge, then a couple of nits.

**Blocking**

1. `src/inspect_robots/eval.py:317`: the `eval()` docstring still says of `before_scoring` that "Exceptions it raises propagate to the caller." After this PR that is no longer true for `SafetyAbort`/`EmbodimentFault` (they are now recorded and the run returns an error log), and it was the sentence #401 cites as the reason the old behaviour was spec-conformant. Please amend it to say that halt-class errors are recorded and halt the run like a rollout fault, and that other exceptions still propagate. The `observe_parked` paragraph at `:309` ("which halt the eval") would read better with the same one-clause clarification, since a reader could still take "halt" to mean "raise".

2. `tests/test_eval_orchestration.py:1118`: the rewritten test proves the eval no longer raises, but not the thing the issue is about. Please strengthen it, or add a sibling, so that it would catch a regression of the log-loss itself:
   - assert the log reached disk (`assert list(tmp_path.glob("*.json"))`, as `test_halted_eval_with_pass_at_k_reducer_still_writes_log` at `:393` does), and
   - use two or three epochs with a park hook that only faults on the second call, then assert the first trial's scores survive and no third trial runs: e.g. `log.samples[0].epochs == ({"success_at_end": ...}, {})`, `log.samples[0].status == "error"`, `log.results.errored_trials == 1`, `len(log.samples[0].termination_reasons) == len(log.samples[0].epochs)`, and `embodiment.park_calls == 2`. That is the "trial 40/40 loses all 40" scenario from #401 in miniature, and it also pins the parallel-tuple alignment your `epoch_dicts.append({}) / judgements.append(None) / notes.append(None)` block is responsible for.

3. There is no test for `before_scoring` itself raising a halt-class error, even though the issue and the PR title both name it. Coverage stays at 100% because the same `except` serves both hooks, so nothing currently guards that path. A short parametrized test next to `test_before_scoring_exception_propagates` (`:995`) with a hook that raises `SafetyAbort`/`EmbodimentFault` and asserts `log.status == "error"` plus the on-disk file would do it, and it would also lock in that the `RuntimeError` case at `:995` keeps propagating.

4. `CHANGELOG.md:10`: please add an entry under `[Unreleased]` / `### Fixed` in the house style, e.g. "**Core:** a `SafetyAbort`/`EmbodimentFault` raised by `observe_parked()` or the `before_scoring` hook now halts the run through the same path as a rollout fault, so the JSON log of every completed trial is written instead of discarded ([#401](https://github.com/robocurve/inspect-robots/issues/401))."

**Nits (non-blocking)**

- `src/inspect_robots/eval.py:608-617`: the `epoch_scores` buffer and the "commit only after all scorers succeed" loop do not change anything observable. A scorer raising a non-halt exception still propagates out of `_run_eval` and the log is lost exactly as before, so the partial `per_scorer_scores` were never read; and a scorer raising `SafetyAbort` is not something the issue asks for and is untested. I would drop the buffering and narrow the `try` to the hook phase (park + `before_scoring`), which also removes the 40-line re-indent of the `observe_parked` block and makes the diff much easier to read. If you do want scorers to be able to halt the run, say so in the docstring and add a test; otherwise keep it out of this PR.
- `src/inspect_robots/eval.py:627-644`: the new handler re-does the bookkeeping that the `record.status != "success"` branch at `:560-567` already does (`epoch_dicts.append({})`, `errored_trials += 1`, `judgements`/`notes`). If you restructure so the hook phase runs first and, on a halt, only mutates `record.status`/`record.error` and the halt variables, the existing branch can handle the rest and there is one place that knows what an unscored trial looks like.
- The branch carries a merge commit from upstream (d5c78eae). Not a problem for a squash merge, but a rebase would give a cleaner history if you are pushing again anyway.

Amresh-01:
Hi @jeqcho - PR ready for review.

Amresh-01:
Hi @jeqcho PR ready for review

Amresh-01:
Hi @jeqcho Pr ready for Review.

PR #405, opened 2026-08-22
docs(frames): define pre/post side-car contract

Vedangalle:
## Summary

Follow up on the non-blocking documentation points from the review of #209.

The frame-store behavior was implemented and tested, but two public descriptions still reflected the old pre-action-only model. This PR makes the migration and indexing contract explicit:

- move the changelog entry from "Added" to "Changed" and name the migration from `step.result.observation.images` to `step.result_image_refs`;
- document reset frame `0`, post-action frame `t + 1`, and the resulting `n + 1` files for a camera present throughout an `n`-step trial;
- distinguish `image_refs` (pre-action) from `result_image_refs` (post-action);
- record that a first-decision failure can leave reset frames for forensics without producing a `StepRecord`;
- correct `default_fps` documentation: adjacent stored frames span one control interval even though the sequence contains an extra reset frame.

No runtime behavior or public interface changes.

## Verification

- [x] `pre-commit run --all-files`
- [x] `pre-commit run --all-files --hook-stage pre-push`
- [x] Ruff lint and format checks
- [x] Strict mypy: 86 source files
- [x] Pytest: 1,619 passed, 6 skipped
- [x] Coverage: 100% line and branch
- [x] Generated API reference
- [x] Docusaurus production build

Follow-up to #209.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,15 @@ All notable changes to this project are documented here. The format is based on
 
 ### Changed
 
+- **Core:** when `FrameStore` is active, both the pre-action and post-action
+  observations recorded in each `StepRecord` now omit inline camera arrays.
+  Consumers that previously read terminal images from
+  `step.result.observation.images` must load `step.result_image_refs` instead;
+  `step.image_refs` remains the pre-action mapping. A trial with `n` completed
+  steps stores the reset frame at index `0` and each post-action frame at
+  `t + 1`; a camera present throughout therefore produces `n + 1` files
+  ([#209](https://github.com/robocurve/inspect-robots/pull/209)).
+
 - **Core:** the `--epochs` below-1 guard in `run` and `eval-set` now lives in
   one shared helper, and its error reads `--epochs must be >= 1, got 0`
   (naming the offending task in `eval-set`) instead of doubling the wording
@@ -229,11 +238,6 @@ All notable changes to this project are documented here. The format is based on
   `messages`); construction guards now diagnose explicit wire conflicts,
   Messages endpoint routing mistakes, and possible silent tool drops on an
   explicit Chat Completions endpoint (plan 0044, #278).
-
-- `FrameStore` now persists each post-action observation once and exposes it
-  through `StepRecord.result_image_refs`. Stored records strip camera arrays
-  from both pre-action and post-action observations, and the terminal visual
-  state is recoverable for offline scoring.
 
 - The Rerun sink now sends a per-trial blueprint that groups labeled action
   dimensions by arm, overlays aligned measured state, and lays out cameras,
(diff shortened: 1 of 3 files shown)


jeqcho (approved):
Thanks @Vedangalle — this is exactly the follow-up I asked for on #209, and it lands each point cleanly.

What I especially like: you didn't just paraphrase the implementation, you stated the contract in terms a consumer can act on. The "reset at `0`, result of step `t` at `t + 1`, `n + 1` files for a camera present throughout" phrasing is precise, and the "camera present throughout" qualifier quietly handles the cases where a camera drops out mid-trial. Naming the migration path (`step.result.observation.images` → `step.result_image_refs`) in the CHANGELOG is the thing external scorers actually need, and moving it from Added to Changed within `[Unreleased]` is the right call rather than a rewrite of released history. The first-decision-failure paragraph turns my "tiny observation" into a documented, intentional forensics feature — good.

What I verified against the code:

- `src/inspect_robots/rollout.py:308` stores the reset observation at index `0` before `controller.next_action` is first called (`:366`), so a step-0 policy failure leaves reset frames on disk with `record.steps` empty — matches the new paragraph in `docs/guide/logging-and-rerun.md`.
- `rollout.py:439-441` stores `result.observation` at `t + 1`, and `:477-479` carries `result_obs_rec`/`result_refs` forward as the next step's pre-action pair, so each frame is written exactly once — `n + 1` files per camera, as documented.
- `rollout.py:209-215` (`_store_frames`) returns `(obs, None)` with no store, and `replace(obs, images={})` plus refs with one, so "images empty while a store is active; both ref mappings `None` without one" is accurate.
- `grader.py:111` already reads `result_image_refs if final else image_refs`, consistent with the pre/post split the guide now describes.
- `default_fps`: each rollout `t` is one `embodiment.step(action)` call, so adjacent stored frames really are one nominal control interval apart; the new docstring is correct where the old "once per rollout step" was off by one.
- The `/api/#inspect_robots.rollout.StepRecord` link matches the anchor style used elsewhere in that guide; no other doc/README/CLAUDE.md still carries the old per-step wording.
- CHANGELOG diff confirmed: removed lines were under `[Unreleased] › Added`, new entry under `[Unreleased] › Changed` with the #209 link.

Nits (none blocking — take or leave):

1. `src/inspect_robots/_video.py:121-122` — the rewrap left an orphan line ("literal, and a hand-edited" / "`control_hz` must fall back…"). Ruff doesn't reflow docstrings so it passes, but joining those two lines would read cleaner.
2. `docs/guide/logging-and-rerun.md:184-185` — "Without a frame store … both ref mappings are `None`" is true, but a reader may invert it into "`None` implies no frame store". Per `_store_frames`, a store that is active on a camera-less observation also yields `None`. A short parenthetical ("also `None` for an observation with no cameras") would close that gap.
3. `CHANGELOG.md:38-46` — the move dropped the old entry's one-line "why" (terminal visual state now recoverable for offline scoring; memory bound preserved). Purely optional, but a half-sentence of motivation alongside the migration instruction helps readers decide whether the change affects them.

Merging as-is; happy to take the nits in a later pass if you feel like it.

PR #409, opened 2026-08-27
feat(rollout): symmetric input-side Perturber protocol and perturbation events (#403)

### LINKED ISSUE #403: Symmetric input-side hook alongside Approver, plus environment provenance in EvalLog
The Approver protocol sits between the policy and the embodiment and can pass, clamp or veto an action. There is no symmetric hook on the way in. Today a robustness sweep means forking the embodiment to mutate the observation or the instruction, which breaks the embodiment-agnostic Task abstraction the README sells.

Would you take a PR for a Perturber protocol that sits between embodiment.observe() and policy.act(), returning a possibly-modified Observation and instruction, with the applied perturbation id and parameters recorded on the TrialRecord? Same shape as Approver, opposite direction. It stays a no-op by default so nothing changes for existing users.

Second, smaller ask. EvalLog records the drawn seed and handles seed=None well, but it records nothing about environment provenance. Two logs with the same seed and the same task name are not comparable across releases if the sim build, the asset revision or the policy checkpoint moved. A provenance block in the schema-versioned EvalLog (sim build id, asset revision hash, policy checkpoint hash) would make a published number re-derivable months later.

Happy to open either as a PR. Asking first so I do not send you something you did not want.

Adityakk9031:
Summary: Implements the input-side Perturber protocol and AutoPerturber in approver.py, perturbation_event in transcript.py, and rollout integration. Fixes part of #403. cc @jeqcho

close https://github.com/robocurve/inspect-robots/issues/403

### DIFF
--- a/src/inspect_robots/approver.py
+++ b/src/inspect_robots/approver.py
@@ -17,7 +17,7 @@ import numpy.typing as npt
 
 from inspect_robots.errors import SafetyAbort
 from inspect_robots.spaces import ABSOLUTE_CONTROL_MODES, Box, RotationRepr
-from inspect_robots.types import Action
+from inspect_robots.types import Action, Observation
 
 
 @runtime_checkable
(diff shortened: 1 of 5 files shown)


Adityakk9031:

Issue #408, opened 2026-08-26
agent: put the action bounds in the system prompt and give a stall/at-bound flag in move results


jeqcho:
Today the only place the model learns the Cartesian bounds is the `move_to` tool description ("Per-dimension bounds: left_x: [0.15, 0.48], ..."), and a clamped or unreachable target comes back as the same "executing move_to over N steps" as a normal move. Agents then infer contact or reach limits from stalls and get it wrong (2026-08-26, rig-2: 18 calls attributing a `left_y` bounds wall to arm reach; rig-3: a correct dead-band computation, but only after 17 calls).

Proposal:
1. Repeat the bounds (and which axes are pinned) in the system prompt's embodiment section, next to the frame conventions, so they are part of the model's initial world model rather than a tool footnote.
2. Have the move result carry a short status: `clipped: left_y 0.30 -> 0.25`, `stalled after 12 steps (reached x=0.148)`, or `ok`, computed from the approver record and the readback vs. target, so the model does not have to diff `eef_state` itself.
3. Make the `give_up` hint say that operators can change limits between trials (the rig-3 operator offered exactly that mid-run and the model did not know it was an option).

Related: robocurve/inspect-robots-yam#135 and the yam issues on default bounds and `bounds_clamped` reporting filed today.

Issue #407, opened 2026-08-26
agent: tell the model the env step budget, not only the LLM-call budget


jeqcho:
The agent system prompt says "You have a budget of {N} LLM calls for the whole trial" but says nothing about `max_steps`. Under `eef_pos` a single `move_to` costs 30–65 env steps (a 20 cm move at the default `max_speed_frac` ≈ 4 s at 10 Hz), so with the rigs' `max_steps = 1200` a trial can end on steps with most of its LLM calls unused. On 2026-08-26 a Jenga run on rig-2 (`adhoc_02b38d64`) hit the 1200-step cap at 65 calls with `max_llm_calls = 100`; the transcript ends mid-plan with no `done`/`give_up`, and the model had spent 148 steps (12%) on look-around and calibration moves.

Proposal: include the step budget and the cost of a typical move in the system prompt (the embodiment's `bind_task` already knows the horizon), and report steps used / remaining in each observation alongside the call count so the model can pace itself. Optionally warn in the tool result when a single move consumed more than ~5% of the budget.

PR #411, opened 2026-08-27
feat(agent): surface action bounds in system prompt and report target reached (part of #408)

Adityakk9031:
### Summary

Part of #408. Surfaces action bounds and pinned axes in the system prompt, updates \give_up\ description to note workspace limits, and reports \Target reached.\ when residual motion is within tolerance.

### Changes
- **System prompt:** Appends \Embodiment bounds: <bounds_text>\ and lists \Pinned dimensions: ... (fixed; do not attempt to move them)\ in \LLMAgentPolicy.reset()\.
- **Toolset:** Adds \ounds_text\ and \pinned_labels\ properties checking \self._step_limits[index] == 0\ with \strict=True\.
- **Tool description:** Updates \give_up\ description to instruct model to mention workspace limits when giving up.
- **Move narration:** Replaces offset narration with \Target reached.\ when residual magnitude is within \_TARGET_REACHED_TOLERANCE\ (\1e-4\).
- **Tests & Docs:** Added unit tests in \	est_tools_motion.py\ and e2e tests in \	est_policy_e2e.py\ for prompt bounds and \Target reached.\ narration. Updated \CHANGELOG.md\ and \plugins/inspect-robots-agent/README.md\.

cc @jeqcho


Adityakk9031:

jeqcho (changes requested):
Thanks @Adityakk9031 for picking up #408 — this is one of the issues with the clearest field evidence behind it, so I'm glad to see movement. A few things are genuinely good here: `pinned_labels` is the right abstraction (the model needs to know which axes are walls, not just the numbers), the `give_up` hint is a faithful reading of item 3 in the issue, and the toolset unit test pins down the exact `bounds_text` format and the pinned-axis detection with a fixed dimension, which is the case that matters. Lint, format, and mypy are all clean.

The problem is that the PR title says "Closes #408" but two of the issue's three asks are not implemented, and the piece that is implemented (the properties) is never consumed. Details below.

### Blocking

1. **The bounds never reach the system prompt** (`plugins/inspect-robots-agent/src/inspect_robots_agent/_tools.py:136-148`, `policy.py:761-770`). `bounds_text` and `pinned_labels` are defined but nothing in `policy.py` reads them; the model still only sees the bounds as a footnote in the `move_to` description, which is exactly the situation #408 item 1 describes. The wiring is small: `bind()` stores `self._toolset` at `policy.py:750`, and `reset()` formats the prompt right after at `policy.py:764-765`, so you can append an embodiment-bounds sentence (and the pinned axes, if any) to `formatted` there, next to where `Embodiment notes:` is attached. Please add an e2e assertion that `script.requests[0]["messages"][0]["content"]` contains the bounds line and the pinned labels, which is the test that would actually fail without the fix.

2. **No `clipped` / `stalled` / `ok` status in the move result** (issue item 2). The only narration change is `" Target reached."` when the residual is at most `1e-4` (`policy.py:1159-1165`); the playout-count narration the PR title mentions already existed on main. A clamped or stalled move still reads the same as a normal one to the model. The ingredients are already in the observation: the rollout attaches `extra["approvals"]` as `{"t": step, "detail": "clamped" | "delta_clamped"}` records for the window since the last policy call (`src/inspect_robots/rollout.py:356`, `:420-424`), and `_approvals_line` at `policy.py:1202` already aggregates them per observation. Combining those with `pending.issued_step` / `pending.chunk_len` and the residual gives you `clipped`, `stalled after N steps (reached x=...)`, or `ok` without touching the core. If you would rather scope this PR to items 1 and 3 only, that is fine by me, but then change the body to "Part of #408" so the issue stays open for the status flag.

3. **The new branch is not asserted anywhere.** No test file contains the string `Target reached`. Per-test coverage contexts show `policy.py:1165` is only reached incidentally by `test_chained_capture_uses_neutral_narration_when_step_does_not_advance` and `test_missing_queued_camera_is_narrated_without_raising` (both happen to land exactly on the target), and neither checks the narration for it. Add a case where `q` lands on the target and assert `"Target reached."` is present and `"remaining offset"` is absent; `test_chained_capture_arrives_on_next_observation_with_playout_and_residual` at `tests/test_policy_e2e.py:2258` is the natural place for a sibling.

4. **CHANGELOG and version bump.** The root `CHANGELOG.md` records plugin changes under `## [Unreleased]` as `**Agent plugin (x.y.z):** ...` entries, and plugins keep static versions that are bumped in the PR (`plugins/inspect-robots-agent/pyproject.toml` is at 0.25.0; this is a prompt and tool-surface change, so 0.26.0). Please add both, and once the bounds are in the prompt, update the system-prompt paragraph in `plugins/inspect-robots-agent/README.md` around lines 164-185 so the docs match.

### Nits (not blocking)

- `_tools.py:146`: `pinned_labels` decides "fixed" with `lo == hi`, while the toolset's own notion of fixed is `self._step_limits[index] == 0` (`_tools.py:450`, and the `movable = high64 > low64` mask in `build_toolset`). Reusing `_step_limits` keeps the two from drifting. Also prefer `strict=True` in that `zip`, as `_move_absolute` does; the three arrays are the same length by construction.
- `policy.py:1159`: `1e-4` is a magic number in native action units (0.1 mm for `eef_pos`, 0.1 mrad for joints). A named module constant with a one-line comment on units would help the next reader.
- The `give_up` description text (`_tools.py:236-240`): the parenthetical "(Note: ...)" reads as an aside. The point of item 3 is that the model should name the limit as the blocker so the operator can act, so something like "If a workspace limit is what stops you, say so in your reason: operators can widen limits between trials." carries the intent more directly.
- The new test is a pure `Toolset` unit test with function-local imports; `tests/test_tools_motion.py` holds the other toolset tests with top-level imports and would be a more natural home.
- The first commit references #407 (step budget) but the final diff carries none of it, and the PR title/body describe playout-status reporting that predates this branch. Tightening the title and body to what the diff does will make the changelog entry easier too.

Happy to re-review once the prompt wiring and a real narration test are in; the shape of the change is right, it just stops one step short of the model actually seeing it.

Adityakk9031:

1. **System prompt wiring:** \
eset()\ now formats \Embodiment bounds:\ and lists \Pinned dimensions: ... (fixed; do not attempt to move them)\ using \self._toolset\. Added e2e test \	est_system_prompt_contains_embodiment_bounds_and_pinned_labels\ in \	est_policy_e2e.py\.
2. **Narration:** Defined named constant \_TARGET_REACHED_TOLERANCE = 1e-4\ (0.1 mm for \eef_pos\, 0.1 mrad for joints). In \_pending_narration\, if magnitude is within tolerance, narrates \Target reached.\ without remaining offset. Added sibling test \	est_chained_capture_arrives_on_target_reports_target_reached\ in \	est_policy_e2e.py\.
3. **Toolset & Give up:** In \pinned_labels\, check \self._step_limits[index] == 0\ with \strict=True\. Updated \give_up\ description to note workspace limits directly. Moved toolset unit test to \	ests/test_tools_motion.py\.
4. **Docs & Changelog:** Added entry in \CHANGELOG.md\ under \[Unreleased] -> Added\, updated \plugins/inspect-robots-agent/README.md\, retitled PR to 'part of #408', and rebased onto \main\.

Adityakk9031:

1. **System prompt wiring:** `reset()` formats `Embodiment bounds:` and lists `Pinned dimensions: ... (fixed; do not attempt to move them)` using `self._toolset`. Added e2e test asserting the bounds and pinned labels in the system prompt.
2. **Target reached narration:** Added named constant `_TARGET_REACHED_TOLERANCE = 1e-4` (0.1 mm / 0.1 mrad). When residual is within tolerance, reports `Target reached.` without remaining offset. Added sibling test `test_chained_capture_arrives_on_target_reports_target_reached`.
3. **Toolset & Give up:** In `pinned_labels`, checked `self._step_limits[index] == 0` with `strict=True`. Updated `give_up` description to clearly note workspace limits. Moved toolset unit test to `tests/test_tools_motion.py`.
4. **Docs, Changelog & CI:** Added `CHANGELOG.md` entry under `[Unreleased]`, updated `plugins/inspect-robots-agent/README.md`, resolved line-length formatting, and retitled PR to 'part of #408'. All 525 tests and all 20 CI jobs are green (`ci-ok pass`).

PR #412, opened 2026-08-27
feat(agent): surface environment step budget in system prompt and observations (#407)

### LINKED ISSUE #407: agent: tell the model the env step budget, not only the LLM-call budget

Adityakk9031:
Summary: Surfaces the environment step budget alongside the LLM call budget in the system prompt, adds bind_task hook to receive the trial horizon from the task envelope, and reports steps used/remaining in observations and echo output. Closes #407. cc @jeqcho


Adityakk9031:

jeqcho (changes requested):
Thanks @Adityakk9031, this is a clean read of #407 and a nice small diff. Things I liked specifically:

- The core hook lands exactly where plan 0013 said a policy-side `bind_task` should go: duck-typed via `getattr`, right after the embodiment hook, before any rollout, so a raising hook aborts before hardware is touched. No Protocol change, no `__all__` change, existing policies untouched.
- `_image_parts` still calls `_step_label(observation)` without `max_steps`, so the camera-frame label join key stays `step N`. That was the one place a "helpful" change would have broken the report, and you left it alone.
- `remaining = max_steps - env_step` is exactly right given the rollout's `while t < max_steps` loop with `env_step = t`.
- Both tests fail without the change, and the core test keeps the 100% gate (I ran it: 1488 passed, 100.00%).

A few asks before merge. First three are blocking, rest are nits.

**Blocking**

1. `CHANGELOG.md` under `## [Unreleased]` needs two bullets: a **Core:** one for the new optional policy `bind_task(envelope)` hook (link #407), and an **Agent plugin (0.26.0):** one for the step budget in the system prompt and observations. Plugin behaviour changes also bump the static version in `plugins/inspect-robots-agent/pyproject.toml:7` (currently `0.25.0`); that is how the plugin gets republished at the next release.

2. The hook is a new core contract, so document it the way plan 0013 did for the embodiment side:
   - `src/inspect_robots/policy.py:58` says "five optional hooks" and lists them; add `bind_task(envelope)` there (when it fires, that it never fires on a direct `rollout()`, so policies must keep a fallback, which yours does).
   - Add a no-op default on `PolicyBase` next to `bind` (`src/inspect_robots/policy.py:106`), mirroring `EmbodimentBase.bind_task` at `src/inspect_robots/embodiment.py:152`. It needs one covering test line to keep the core gate at 100%.
   - Update the `policy.py` and `eval.py` rows in `src/inspect_robots/CLAUDE.md:14` / `:28`.
   - The comment at `src/inspect_robots/eval.py:387` could say "plan 0013 anticipated this" so the next reader sees why the name is `bind_task` and not `bind`.

3. `plugins/inspect-robots-agent/src/inspect_robots_agent/policy.py:779-781` hard-codes "a typical move takes ~10-60 steps". That figure is only true at 10 Hz with the default `max_speed_frac`; the toolset caps a move at `_MAX_DURATION_S * control_hz` steps (`_tools.py:38,111`), so at 30 Hz a single move can be 300 steps and the prompt would be telling the model it has 5x more moves than it does. Two options, either is fine: (a) derive the sentence from the bound toolset's hz and cap (something like "a single move can take up to N steps (10 s at H Hz)"), or (b) drop the number and instead point the model at the per-move cost it already receives: every move's tool result says `executing move_to over N steps (X.Xs)` (`_tools.py:530-532`). I would lean (b) plus a short "pace yourself against the step budget, not only the call budget" clause, since that is the failure mode in #407.

**Nits**

4. `policy.py:1266`: `max(0, ...)` cannot trigger (`env_step < max_steps` by construction). Harmless, but it invites the reader to look for a case that does not exist; a plain subtraction reads truer.

5. The echo path (`step 10/200` at `policy.py:910`) has no assertion. One `capsys`/echo check in the new test would pin it; `test_policy_e2e.py` already has echo tests to copy from.

6. `plugins/inspect-robots-agent/README.md:164-181` describes what the system prompt contains; add a line for the new `Environment step budget:` section so users know it appears only under `eval()`, not under a bare `rollout()`.

7. Optional, and fine to leave for a follow-up: #407 also floated a warning in the tool result when one move eats more than ~5% of the budget. Not needed for this PR.

Once 1-3 are in I am happy to merge; the core of the change is right.

Adityakk9031:

1. Added \CHANGELOG.md\ entries under \[Unreleased] -> Added\ for both Core (\ind_task\) and Agent plugin (0.26.0), and confirmed the plugin version is \

Adityakk9031:

1. **Changelog & version:** Added bullets in `CHANGELOG.md` under `## [Unreleased]` for Core (`bind_task(envelope)`) and Agent plugin (0.26.0), and bumped the version in `plugins/inspect-robots-agent/pyproject.toml` to `0.26.0`.
2. **Core hook contract:** Documented `bind_task` in `src/inspect_robots/policy.py`, added no-op default on `PolicyBase`, updated `CLAUDE.md`, and referenced plan 0013 in `eval.py`.
3. **Step budget wording:** Updated system prompt guidance to instruct the agent to pace itself against the step budget and per-move cost rather than hard-coding step numbers.
4. **Plain subtraction:** Used plain subtraction `max_steps - env_step` in `policy.py`.
5. **Echo assertion:** Added `capsys` test asserting the `step 10/200` echo output.
6. **Documentation & CI:** Updated `plugins/inspect-robots-agent/README.md` and formatted all tests. All 523 tests and all 20 CI jobs are green (`ci-ok pass`).

jeqcho (approved):

PR #415, opened 2026-08-29
feat(scorer): make the affirmative operator-verdict contract public API

### LINKED ISSUE #414: [feature] Make the affirmative operator-verdict contract public API
**Is your feature request related to a problem?**

On real hardware there is no success oracle, so an attended run's ground truth is the string a human types. The vocabulary deciding which strings mean success is private:

```python
# src/inspect_robots/scorer.py:236
_OPERATOR_SUCCESS = frozenset({"success", "pass", "yes", "y", "1", "true"})
```

Benchmarks that grade real-world runs need the same answer, and because the set is private they cannot import it. KitchenBench copied it instead:

```python
# kitchenbench/src/kitchenbench/scoring.py:17
# Mirrors inspect_robots.scorer._OPERATOR_SUCCESS (private there) - keep the two in sync.
_AFFIRMATIVE = frozenset({"success", "pass", "yes", "y", "1", "true"})
```

and guards the copy with a test that reaches across repos into the private symbol (`kitchenbench/tests/test_world.py:50`). Two problems with that arrangement:

1. **It pins a private symbol from another repo.** `tests/test_api_snapshot.py` states that anything outside `__all__` or prefixed `_` carries no stability guarantee, and the snapshot fence does not cover private names. Renaming or inlining `_OPERATOR_SUCCESS` is a legal refactor here that breaks KitchenBench's test suite after release, not during review.
2. **The guard checks the vocabulary, not the comparison.** Both sites also do `verdict.strip().lower() in <set>` and handle `None` separately. That half is duplicated and unguarded: the two frozensets can stay equal while the normalization around them drifts, and the equality assertion stays green.

**Describe the solution you'd like**

A public predicate in `inspect_robots.scorer`, exported from `__all__`, that owns the whole contract (vocabulary, case-folding, surrounding whitespace, and the "no judgement recorded" case):

```python
def is_affirmative_verdict(verdict: str | None) -> bool: ...
```

`_OPERATOR_SUCCESS` stays private as its backing data. `_OperatorScorer` calls the predicate instead of restating it, so the repo holds one implementation. KitchenBench then drops `_AFFIRMATIVE` and the private-symbol guard in a follow-up (blocked on a release, since it resolves inspect-robots from PyPI).

No behavior change: existing operator-scorer tests pass unchanged.

**Scope check**

- [x] This belongs in the **framework** (not a specific benchmark or backend).
- [x] If it's a backend (sim/model), it could ship as a separate plugin package. (n/a)

**Alternatives considered**

- **Export the frozenset** (`OPERATOR_SUCCESS_VERDICTS`) instead of a predicate. Smaller surface, but it only shares the vocabulary and leaves the normalization duplicated, which is the half the current guard already misses.
- **Leave it private and keep the guard.** Works until someone refactors the private name, and never covers normalization drift.
- **Move scoring of operator verdicts into the framework entirely.** Larger change and out of scope; benchmarks legitimately want their own scorer semantics (KitchenBench's `task_success` also accepts a `"success"` termination reason).

**Additional context**

Patch is written and green locally (100% coverage, mypy strict, ruff clean on the touched files). Happy to open the PR.

jav359003:
## Summary

On real hardware there is no success oracle, so an attended run's ground truth is the string a human types. The vocabulary that decides which strings count lived in a private `scorer._OPERATOR_SUCCESS`, so KitchenBench copied it (`kitchenbench/scoring.py:17`) and guards the copy with a test that imports the private symbol across repos (`kitchenbench/tests/test_world.py:50`).

That guard is insufficient in two ways:

1. It pins a private symbol from another repo. Per `tests/test_api_snapshot.py`, private names carry no stability guarantee and the snapshot fence does not cover them, so renaming `_OPERATOR_SUCCESS` breaks KitchenBench after release rather than during review.
2. It compares the vocabulary but not the comparison. Both sites also do `verdict.strip().lower() in <set>` plus a `None` check. The two frozensets can stay equal while that normalization drifts, and the assertion stays green.

This promotes the contract to public API so there is one owner for both halves.

## Changes

- Add `is_affirmative_verdict(verdict: str | None) -> bool` to `inspect_robots.scorer`: the vocabulary *and* the comparison rules (case-insensitive, whitespace-tolerant, `None` is not affirmative).
- `_OperatorScorer` now calls it instead of restating the comparison, so the repo holds exactly one implementation.
- Export from `inspect_robots.__all__` and add to the API-snapshot `EXPECTED` set.
- `_OPERATOR_SUCCESS` is unchanged and stays private as backing data.
- Tests: the full vocabulary, case/whitespace variants, rejections (including `"yes please"`, which contains an affirmative word but is not one), `None`, and a consistency test asserting the scorer's value *is* the predicate's, so a second copy cannot be reintroduced silently.
- CHANGELOG entry under Unreleased.

No behavior change. The existing `test_operator_scorer_reads_recorded_verdict` passes unchanged, which is the evidence.

## Follow-up (not in this PR)

KitchenBench drops `_AFFIRMATIVE` and the private-symbol guard in favor of importing this. It is blocked on a release, since KitchenBench resolves `inspect-robots` from PyPI. Happy to open it as soon as a version ships.

## Open question

I chose a predicate over exporting the frozenset because it also owns the normalization that the current cross-repo guard cannot check. If you would rather expose `OPERATOR_SUCCESS_VERDICTS` as data and leave normalization to callers, say the word and I will switch.

## Checklist

- [x] `pre-commit install` done; hooks pass (commit and push stages)
- [x] Tests added/updated
- [x] Coverage stays at **100%** (`pytest --cov`: 1639 passed, 100.00%)
- [x] `ruff check` and `ruff format --check` pass on the touched files
- [x] `mypy` passes (strict), 86 source files clean
- [x] Public API changes reflected in `inspect_robots.__all__` and the API-snapshot test
- [x] Core stays NumPy-only (no new dependencies)

## Related

Closes #414


jeqcho (approved):
Thanks @jav359003, this is a tidy one. The issue write-up does the hard part: it names precisely why the current KitchenBench guard is insufficient (it pins a private name, and it compares the vocabulary but not the normalization around it), and the PR fixes exactly that and nothing else. I checked KitchenBench and the copy is where you say it is, so the motivation is real, not hypothetical. Choosing a predicate over exporting the frozenset is the right call, and I would keep it that way: the `strip().lower()` plus `None` handling is the half that drifts silently, so it belongs behind the function.

What I verified:

- Scope matches #414 exactly: one new public function, `_OperatorScorer` delegates to it, `__all__` and `tests/test_api_snapshot.py:38` updated together, CHANGELOG entry under Unreleased. Nothing unrelated in the diff.
- No behavior change. `src/inspect_robots/scorer.py:249` is the old expression verbatim, and the early `None` return in `_OperatorScorer.__call__` (`scorer.py:260`) is kept so the "no operator judgement recorded" explanation survives. `session.py:697` and `session.py:722` (the console path that writes `"y"`/`"n"` and lower-cases free text) still agree with the vocabulary. No hardware path is touched; this is a pure reader on the recorded string.
- Tests are meaningful: full vocabulary, case and whitespace variants, rejections including `"yes please"` and `"   "`, `None`, and `test_operator_scorer_agrees_with_the_public_predicate` which pins the scorer to the predicate so a second copy cannot creep back in. Ran them under the same install CI uses (`--extra dev`): 1639 passed, 100.00% coverage, ruff and mypy clean.
- `scripts/gen_api_docs.py` lists `inspect_robots.scorer` in the Scoring section and picks up every non-underscore member, so `is_affirmative_verdict` lands on the API reference page with no extra work.

Nits, none blocking, take or leave:

1. `docs/guide/scoring.md:60-62` is where benchmark authors read about operator verdicts. One sentence there pointing at `is_affirmative_verdict` ("benchmarks that read `operator_judgement` themselves should call this rather than restate the vocabulary") would make the new function discoverable before someone reaches for a copy again. Happy to take that as a follow-up.
2. `src/inspect_robots/scorer.py:245-246` uses em dashes in the docstring. The repo's writing-style rule targets prose that ends up on the docs site, which generated API docs technically are. Existing docstrings in this file already use them, so this is consistency with what is there rather than a rule you broke; mentioning it only so you know the rule exists if you touch docs pages later.

Merging once CI is green. Ping me once a release ships and I will look at the KitchenBench follow-up that drops `_AFFIRMATIVE` and the private import.

jav359003:
Thanks for the thorough read, and for checking the KitchenBench side independently.

Docs sentence is up as #418, placed right after the `operator_scorer` line since that is where a benchmark author is already reading when they decide to handle `operator_judgement` themselves.

Noted on the em dashes. I matched the surrounding docstrings in that file, but I will keep the rule in mind on docs pages, and #418 has none.

Will ping you once a release ships and open the KitchenBench follow-up that drops `_AFFIRMATIVE` and the private import.

jav359003:
@jeqcho 0.58.0 shipped today with is_affirmative_verdict in it, so here's the
KitchenBench follow-up: robocurve/kitchenbench#37. Drops _AFFIRMATIVE, drops
the cross-repo private import in test_world.py, floor to >=0.58, relocked.
376 passed, 100% coverage, ruff and mypy clean.

Its checks haven't started, I think because it's my first PR on that repo and
the workflow needs approving.

PR #424, opened 2026-08-31
fix(eval-set): stop after in-trial robot halt

### LINKED ISSUE #423: eval_set continues to the next task after an in-trial SafetyAbort/EmbodimentFault
## Observed

`_run_eval` absorbs a `SafetyAbort` / `EmbodimentFault` raised inside a trial into the task's log (`status="error"`, `halted` set, remaining scenes skipped) and returns normally (`eval.py`, the `halted` branch and the `[log]` return). `eval_set` therefore proceeds to the next task, whose `rollout()` resets the arm that just raised the halt. Only a halt that escapes `eval()` (from `observe_parked`, `bind`/`bind_task`, `embodiment.info`, or `close()`) stops the set.

`errors.py`'s module doc says a robot that raised a safety halt "must never auto-advance to the next scene unattended". Within one task that holds. Across tasks in a set it does not.

This is pre-existing behaviour; #421 (plan 0079) documents it accurately rather than changing it, and surfaced the gap.

## Proposal

After each task, if its log records a halt (any `SceneResult.error` from a `SafetyAbort`/`EmbodimentFault`, or an explicit `halted` marker on `EvalLog` if one is added), `eval_set` stops and returns `(False, logs)` without starting the next task. `_cmd_eval_set` then prints the existing summary and hint. `fail_on_error` is about scoring policy errors and should not gate this.

Design question: should the stop be unconditional (safety default) or opt-out via a new keyword such as `continue_after_halt=False`? Recommended: unconditional, with the operator re-running the remaining tasks explicitly.

## Tests

`eval_set` over two tasks where the first trial raises `SafetyAbort` from the embodiment's `step()`: one log returned, second task's embodiment never `reset()`.

Vedangalle:
## Summary

Stop `eval_set()` immediately after an in-trial `SafetyAbort` or
`EmbodimentFault`, before another task can reset a robot that requires operator
intervention.

The stop condition is now an explicit, backward-compatible `EvalLog.halted`
field. This keeps the completed forensic log available without making
control flow depend on human-facing error text. The marker is authoritative:
`eval_set()` returns `(False, logs)` as soon as it sees one.

## Changes

- persist `halted=True` for robot halts and operator cancellation, while older
  schema-v1 logs without the field continue to read as `False`;
- stop a task set before the next task starts, with no continue-after-halt
  opt-out;
- preserve halt precedence if a registry-owned embodiment also fails during
  `close()`, reporting that cleanup failure as a `RuntimeWarning`;
- leave ordinary task-error continuation, non-halted cleanup failures,
  `fail_on_error`, and outside-trial halt propagation unchanged;
- document the API, CLI behavior, module contract, migration, and design;
- cover both halt types, exact reset count, persistence, cancellation,
  status-independent marker precedence, cleanup-failure precedence, and the
  ordinary success/error/success control path.

## Validation

- `pre-commit run --all-files` passes;
- the pre-push hook passes, including the 100% coverage gate;
- 1,673 tests pass, 6 optional Rerun tests skip;
- coverage is 100% line and branch (6,866 statements, 2,238 branches);
- strict `mypy`, `ruff check`, and `ruff format --check` pass;
- API-reference generation and the Docusaurus production build pass.

## Checklist

- [x] `mypy` passes (strict)
- [x] Public API changes are reflected in `inspect_robots.__all__` and the API-snapshot test

## Related

Closes #423


Vedangalle:
Thanks for clarifying the intended eval_set contract in #423. Closing this because its central stop-after-in-trial behavior conflicts with that decision. I am keeping the separate cleanup-exception precedence finding out of this PR.

Issue #413, opened 2026-08-27
The persisted log cannot distinguish an embodiment-adopted verdict from a VLM-sampled one


repowazdogz-droid:
## Observed

`_VLMGrader.grade` writes `operator_judgement` from two different sources.

When the trial terminated with a reason in `_DEFINITIVE_REASONS`, the embodiment's verdict is adopted verbatim and no model call is made (`grader.py:203-207`). Otherwise the VLM is called and its parsed verdict is written (`grader.py:247-251`). Both paths append an `operator_event` carrying `source="embodiment"` or `source="vlm"`.

Neither source reaches the log. `operator_event` builds `Event(kind="operator", ...)` (`transcript.py:68-72`), and `eval.py:647-656` persists only events where `event.kind == "operator_message"`. `SceneResult.operator_judgements` ends up holding both kinds of verdict with nothing marking which is which. The only incidental signal is `operator_note`, which is set on the VLM path and left `None` on the adopted path.

`EvalSpec` (`log.py:43-61`) also records no grader configuration, so the model id, the run-level rubric, `max_cameras` and `effort` are absent from the log.

## Why it matters

The two kinds of verdict have different reproducibility. An adopted verdict is a pure function of `termination_reason`, which the log already stores, so it re-derives exactly. A VLM verdict does not: `_post_chat` sends `{model, messages, max_tokens}` with no `temperature`, `top_p` or seed (`_chatwire.py:67-69`), so the provider default applies.

When a metric moves between two runs of the same task, there is no way to tell from the logs which verdicts were model samples and which were fixed, so the change cannot be attributed without re-reading the grader source and inferring from whether a note is present.

## Evidence

`cubepick-reach`, 4 scenes at each of `max_steps` 1, 8, 9, 10 and 80, graded by `gemini-2.5-flash` through an OpenAI-compatible base URL:

```
inspect-robots run --task cubepick-reach -T num_scenes=4 -T max_steps=9 \
  --policy scripted --embodiment cubepick --grader vlm \
  -G model=gemini-2.5-flash \
  -G base_url=https://generativelanguage.googleapis.com/v1beta/openai \
  -G api_key_env=GEMINI_API_KEY \
  --no-prompt --store-frames --seed 0 --log-dir logs
```

[...]

PR #425, opened 2026-08-31
log: record the effective grader configuration in EvalSpec

repowazdogz-droid:
## Summary

Follow-up to #413, and the piece #422 deliberately left out. On #413 you wrote that recording grader configuration in `EvalSpec` "is a separate additive change and is not in that PR; happy to take a PR for it if you want to follow up". This is that change.

#422 records *where* each judgement came from. The persisted `EvalSpec` still does not record *which grader was configured* or the configuration governing it: it carries `policy` + `policy_config` and `embodiment` + `embodiment_info`, but nothing about the grader at all. Two consequences:

1. A run whose every trial degraded to ungraded records `judgement_sources` of `None` throughout, which is indistinguishable on disk from `--grader none`.
2. For the `vlm` grader, the model, run-level rubric, camera cap and effort are resolved inside `vlm_grader()` and then held privately on `_VLMGrader`, so two runs graded by different models against different rubrics produce byte-identical `EvalSpec`s and the prompt cannot be rebuilt from the log.

## Changes

Two additive fields on `EvalSpec`, the parallel of the `policy` / `policy_config` pair beside them:

```python
grader: str | None = None
grader_config: dict[str, Any] = field(default_factory=dict)
```

- `grader` is the registry name, the value the `Grader` protocol already documents as identifying the grader in logs. `None` when the run graded nothing.
- `grader_config` comes from an **optional duck-typed `config()` hook**, read with `getattr` + `callable`. It is deliberately not a `Grader` Protocol member: `Grader` is `@runtime_checkable` and `_grading_hook` gates on `isinstance`, which does enforce data members, so adding it to the Protocol would make every existing out-of-tree grader fail that check. This follows the `bind_task` and `connect_session` precedent. A grader without the hook is still named, with `{}` config.
- `_VLMGrader.config()` reports the resolved `model`, `base_url`, run-level `rubric`, `max_cameras` and `effort`.

Values are resolved rather than requested: a `rubric_file` is stored as the text read from it, an omitted rubric as the default that replaced it, and `effort` keeps `None` (field omitted, provider default) distinct from `"none"` (minimum requested), matching the normalization from plan 0075.

`base_url` and `max_cameras` are included alongside the three you named. `AgentPolicyConfig` records `base_url` for the same reason, that a model id alone does not say which provider answered, and `max_cameras` caps how many frames reach the prompt. Happy to drop either if you would rather keep the field to the three.

`eval_set()` now hands the grader itself down per task rather than a bare bound method, which would otherwise have left every `eval_set` log with no grader recorded.

## Notes

- **Per-scene rubric is not duplicated.** `_rubric_for()` lets `scene.metadata["rubric"]` win per scene, and that value is already persisted in `SceneResult.scene_metadata`. `grader_config` holds the run-level rubric only, and the field comment says so.
- **No secret material is persisted.** The API key is absent by construction, and a test asserts the key text appears nowhere in the written file.
- **Existing and out-of-tree graders are unaffected.** No grader has a `config()` hook today, so `callable()` is `False` and nothing runs. Grading behaviour is untouched; `config()` is called once at spec construction.
- **Old logs remain readable.** Additive with defaults, no schema bump; `EvalSpec(**data["eval"])` fills both. Verified through the CLI as well as in tests: `inspect` and `view` both render a log with the fields stripped.
- Live and final logs share one `EvalSpec` instance, so no shared helper was needed here as it was for `judgement_sources`.

Plan: `plans/0081-grader-config.md`.

Not in scope, the same line #422 drew: surfacing grader configuration in `view` or the `summarize` digest.

## Test plan

`tests/test_grader_config.py` covers the resolved-not-requested rubric (including that the recorded text is the one in the prompt), the effort tri-state asserted against the actual request body, the scene-rubric override leaving the spec alone, the absent API key, a grader with no `config()` hook, an ungraded run, the JSON round trip, and `eval_set`. `tests/test_eval_log.py` covers the golden round trip and a v1 log missing both fields; `tests/test_live_log.py` pins live and final agreement.

Reverting `src/` while keeping the new tests fails 22 of them, so they exercise the behaviour rather than the implementation.

```
uv run ruff check .          All checks passed!
uv run ruff format --check . 159 files already formatted
uv run mypy                  Success: no issues found in 87 source files
uv run pytest --cov          1681 passed, 6 skipped, 100% coverage
```

Run against `uv sync --locked --extra dev` so the tool versions match CI.

## Checklist

- [x] Hooks pass (`ruff check`, `ruff format`, `mypy` strict)
- [x] Tests added
- [x] Public API unchanged (`inspect_robots.__all__` untouched; fields on an existing public dataclass)
- [x] Core stays NumPy-only

## Related

Follow-up to #413. Complements #422.


jeqcho:
Merged, thanks @repowazdogz-droid! Recording the resolved grader settings will make results much easier to trace. The tests for credential exclusion and agreement between live and final logs were good details to cover.

repowazdogz-droid:
No problem, happy to help!

PR #431, opened 2026-09-02
fix(logging): sanitize non-finite NumPy floating scalars to prevent json.dump crashes

Adityakk9031:
### Summary

\JsonLogSink._sanitize\ maps non-finite float values (\NaN\, \±inf\) to \None\ (\
ull\ in JSON) so that \json.dump(..., allow_nan=False)\ generates strict RFC 8259-compliant JSON logs.

However, \_sanitize\ previously only matched Python \isinstance(obj, float)\. In NumPy, floating scalar types like \
p.float32\ and \
p.float16\ inherit from \
p.floating\ and evaluate to \False\ for \isinstance(..., float)\.

When trial metadata, custom scorer outputs, or policy logs stored non-finite \
p.float32\ / \
p.float16\ values (e.g. \
p.float32(np.nan)\ or \
p.float32(np.inf)\), they passed through \_sanitize\ unchanged, causing \json.dump(..., allow_nan=False)\ in \on_eval_end\ to raise \ValueError: Out of range float values are not JSON compliant\. This uncaught error prevented the final \.json\ eval log from being written to disk at the end of the run.

### Changes
- \src/inspect_robots/logging/json_log.py\: Extended \_sanitize\ to match \isinstance(obj, (float, np.floating))\ and safely coerce to Python floats.
- \	ests/test_strict_json.py\: Added unit and end-to-end tests for \
p.float32\ and \
p.float16\ sanitization and serialization.

cc @jeqcho

### DIFF
--- a/src/inspect_robots/logging/json_log.py
+++ b/src/inspect_robots/logging/json_log.py
@@ -23,6 +23,8 @@ import uuid
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 if TYPE_CHECKING:
     from inspect_robots.log import EvalLog, EvalSpec
     from inspect_robots.rollout import TrialRecord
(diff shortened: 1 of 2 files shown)


Adityakk9031:
cc @jeqcho for review - all CI checks passed cleanly.

jeqcho:
Merged, thanks @Adityakk9031. Good catch on the NumPy scalar case. Covering it in the JSON sanitization tests should help keep this from breaking final logs again.

PR #443, opened 2026-09-13
fix(cli): safely format None metric values as 'n/a' in run and eval-set summaries

Adityakk9031:
﻿### Summary

Fixes a `TypeError` crash in `inspect-robots run` and `inspect-robots eval-set` summaries when evaluation metrics contain `None` values (e.g. from inconclusive trials, custom scorers, or unscored tasks).

### Root Cause

In `src/inspect_robots/cli.py`:
- Line 1377 (`_print_run_summary`) formatted metric values using `f"{value:.4g}"`.
- Line 1846 (`_print_eval_set_summary`) formatted metric values using `f"{name}={value:.4g}"`.

When a metric's value is `None`, Python raises `TypeError: unsupported format string passed to NoneType.__format__`, causing the CLI to crash right at the end of a run. In contrast, `_cmd_inspect` (lines 2029 & 2033), `_html.py`, and `_html_index.py` all format `None` metric values safely as `"n/a"`.

### Solution

Added a helper function `_format_metric(value: float | int | None) -> str` in `cli.py` that formats numeric values with `:.4g` and returns `"n/a"` when `value is None`. Updated `_print_run_summary`, `_print_eval_set_summary`, and `_cmd_inspect` to use this unified helper.

### Testing

Added unit tests in `tests/test_registry_cli.py`:
- `test_run_summary_formats_none_metric_as_na`: verifies `_print_run_summary` prints `custom_metric: n/a` without crashing.
- `test_eval_set_summary_formats_none_metric_as_na`: verifies `_print_eval_set_summary` prints `custom_metric=n/a` without crashing.


Adityakk9031:
@jeqcho have a look

jeqcho:
Merged, thanks @Adityakk9031! Using the same formatter for both CLI summaries keeps their behavior consistent. The `n/a` output is much more useful than losing the summary to a formatting error.

PR #444, opened 2026-09-13
fix(capx): clip motion queue targets and waypoints to action space bounds

Adityakk9031:
﻿### Summary

Fixes an issue in `inspect-robots-capx` where `MotionQueue` failed to clamp targets and interpolated motion waypoints to the configured action space `[low, high]` bounds, leading to out-of-bounds actions and cursor desynchronization with physical robots.

### Root Cause

In `plugins/inspect-robots-capx/src/inspect_robots_capx/_motion.py`:
- `MotionQueue.__init__` validates and stores `self._low` and `self._high` from the embodiment's `action_space`.
- However, `_queue_target` never used these bounds: when model-generated code called `move_to_joints(...)` or `_queue_gripper(...)` with values outside the joint limits, unclipped actions were generated and appended to `self._actions`.
- Furthermore, `self._cursor = target.copy()` recorded the out-of-bounds target directly. Downstream safety approvers or embodiments clamped the actions on hardware, causing `self._cursor` to permanently diverge from the physical robot position for all subsequent motion calls.

### Solution

In `_queue_target`:
1. Clamped incoming `target` to `[self._low, self._high]` upfront using `np.clip(target, self._low, self._high)`.
2. Protected each interpolated action waypoint against floating-point boundary overrun via `np.clip(..., self._low, self._high)`.
3. Guaranteed that both `self._actions[-1]` and `self._cursor` strictly adhere to the bounded action space.

### Testing

Added unit tests to `plugins/inspect-robots-capx/tests/test_motion.py`:
- `test_target_exceeding_bounds_is_clipped_to_box`: confirms joint targets outside space bounds generate clipped actions and clamp `self.cursor`.
- `test_gripper_exceeding_bounds_is_clipped_to_box`: confirms gripper targets beyond `high` clamp to the valid bound.
All 13 tests in `test_motion.py` pass.


Adityakk9031:
@jeqcho have a look

jeqcho:
Thanks @Adityakk9031, merged. Keeping the cursor endpoint consistent with the clipped target was a good detail to cover, along with the waypoint bounds.

PR #447, opened 2026-09-14
▚▚ fix(cli): keep the task's epoch reducer under --epochs

breken-ai:
`--epochs N` is documented (`docs/guide/cli.md`, the flag's help text) as overriding the task's epoch count. It also, silently, overrode the task's epoch reducer: a task declared with `Epochs(count=5, reducer="pass_at_2")` (the exact example in `docs/guide/scoring.md`), or with `max`/`mode`, was reduced with `mean` whenever `run --epochs` or `eval-set --epochs` was used. The log looks ordinary; only the number is wrong.

## Root cause

`_apply_epochs_or_exit` in `src/inspect_robots/cli.py` did `replace(task, epochs=epochs)` with a bare `int`. `Task.epoch_spec` turns an int into `Epochs(count=N)`, whose default is `reducer="mean"`, so the task's own `Epochs` object was thrown away, reducer included. Both `_cmd_run` and `_cmd_eval_set` route through this helper.

## Fix

- `src/inspect_robots/cli.py`: override only the count, `replace(task, epochs=replace(task.epoch_spec, count=epochs))`, so the declared reducer carries through. `Epochs.__post_init__` still raises `ConfigError` for a count below 1, so the guided `--epochs must be >= 1, got 0` message (and the `eval-set` task-attributed variant) is unchanged; those pinned tests still pass.
- `CHANGELOG.md`: entry under Unreleased / Fixed.

## Tests

`tests/test_registry_cli.py`:
- `test_apply_epochs_keeps_the_task_reducer`: a task with `Epochs(count=5, reducer="pass_at_2")` patched to 3 epochs must yield `Epochs(count=3, reducer="pass_at_2")`. Without the fix: `reducer: 'mean' != 'pass_at_2'`.
- `test_cli_run_epochs_override_keeps_the_task_reducer`: a registered task with `reducer="max"` and a scorer returning the epoch index, run with `--epochs 2`; the metric must be `1.0`. Without the fix it is `0.5` (mean).

Full suite: 1722 passed, 100% line+branch coverage; ruff, ruff format, and mypy strict clean.

## Not changed

`--epochs` remains a count-only override; there is no flag to select a reducer from the CLI, and this PR does not add one.

---
▚▚ **Shipped by breken** — self-healing software. This one's on us. [breken.ai](https://breken.ai)


jeqcho:
Thanks @breken-ai, merged! The regression tests catch the wrong score as well as the changed reducer, which makes the impact of this bug clear. Nice focused fix.

PR #449, opened 2026-09-14
agent: retry server-side status=failed payloads on the responses wire

### LINKED ISSUE #448: agent: responses wire treats OpenRouter server_error (HTTP 200, status=failed) as fatal instead of retrying
## What happened

A live rig run with `-P wire=responses -P model=deepseek/deepseek-v4.1-flash` through OpenRouter (`base_url=https://openrouter.ai/api/v1`) died on its 12th LLM call:

```
PolicyError: LLM response failed — stream closed with reason: error
```

The wire capture shows the request got HTTP 200 with a Responses object in a terminal failed state that OpenRouter synthesised when the upstream provider dropped the generation stream mid-way (after 2176 reasoning tokens, 46 s):

```json
{"status": "failed",
 "error": {"code": "server_error", "message": "stream closed with reason: error"},
 "error_type": "server",
 "output": [{"type": "reasoning", "status": "completed", ...}]}
```

`ResponsesClient.complete()` retries transport errors, 429 and 5xx up to `max_retries`, but a 200 body with `status: "failed"` goes straight to `_parse_response`, which raises on the first attempt (plan 0022: "failed status fails fast, no retry ... the deliberate, message-bearing terminal state of a request the server accepted, unlike a 5xx where the server never answered").

That reasoning holds for failures like `reasoning token limit reached`, which are deterministic for the same input. It does not hold for `error.code == "server_error"` (or OpenRouter's `error_type: "server"`): that is a 5xx wearing a 200. On the chat wire the same provider hiccup surfaces as an HTTP 5xx and is retried; on the responses wire it ends the trial.

## Proposal

In `ResponsesClient.complete()`, when a 200 payload has `status: "failed"` and the error is server-side (`error.code in {"server_error", "rate_limit_exceeded"}` or `error_type == "server"`), treat it exactly like a 5xx: record it as `last_error`, back off, and retry, never populating the raw-item cache. All other failed payloads keep the fail-fast path from plan 0022 (existing test `test_failed_status_raises_once_and_does_not_populate_cache` stays as is).

Related: #441 (expose `max_retries`/`backoff_s`).

jeqcho:
Closes #448.

On `wire=responses`, OpenRouter reports an upstream provider stream drop as HTTP 200 with `status: "failed"`, `error.code: "server_error"`, `error_type: "server"`. `ResponsesClient` treats every failed payload as terminal (plan 0022) and the trial ends on attempt 0, while the chat wire gets an HTTP 5xx for the same event and retries.

This narrows plan 0022: a failed payload whose error is server-side (`server_error`, `rate_limit_exceeded`, or OpenRouter's `error_type: "server"`) is retried with the existing backoff and never populates the raw-item cache; every other failed payload keeps failing fast with its message. Plan: `plans/0082-responses-failed-server-error-retry.md`.

Plugin version 0.26.0 to 0.26.1 (with `uv.lock`).


jeqcho:
Opened by mistake ahead of agreement on the approach; closing.

